PR #1, opened 2020-01-29
Feat/basic

giovannic:
Create a basic simulation with humans and mosquitos.

* For the time being, humans stay in the Susceptable state. Mosquitos grow from egg > early larvae > late larvae > pupal > susceptable.
* Humans' age variable is updated every 365 timesteps
* There is no infection process yet (later pull request)
* There is no mortality / seasonality yet (later pull request)

NOTE: It looks like there are a lot of files but there is no need to review the "docs" or "man" files. They are automatically generated from my code comments.


R/individuals.R
```diff
@@ -0,0 +1,66 @@
...
+mosquito_limit <- 100 * human_population
+
+#' @title Define model states
+#' @description
+#' create_states creates the human and mosquito states for the model
+create_states <- function() {
+  list(
+    # Human states
+    S       = individual::State$new("S", human_population),
+    I       = individual::State$new("I", 0),
```
bobverity: Trying to match these with my internal map of states. Is "I" the liverstage, i.e. exposed state that leads on to diseased ("D") or treated ("T"), or have I got this wrong? Also, I can't see a prophylactic state.

R/processes.R
```diff
@@ -0,0 +1,156 @@
...
+    ),
+    # Asymptomatic Progression
+    create_fixed_probability_state_change_process(
+      individuals$human,
+      states$D,
+      states$A,
+      parameters$rd
+    ),
+    # Subpatient Progression
+    create_fixed_probability_state_change_process(
```
bobverity: Is there super-infection in this model, i.e. is it possible to go from "U" back to "A", or are flows currently only in one direction?

R/processes.R
```diff
@@ -0,0 +1,156 @@
...
+#' states and variables change over time.
+#'
+#' It lists processes from `infection.R`, `mosquito_emergence.R` and
+#' `mortality.R`; and then exposes them to the model through
+#' `bind_process_to_model`
+#' @param individuals, a list of individuals in the model
+#' @param states, a list of states in the model
+#' @param variables, a list of variables and constants in the model
+#' @param parameters, a list of model parameters
+create_processes <- function(individuals, states, variables, parameters) {
```
bobverity: So I can see how the following processes can be used to define progression through infected states, but I can't see how humans get infected in the first place, i.e. where is the process that takes "S" to the latent stage? Again, apologies if I've missed the point here, still getting up to speed with the method.

bobverity: Just read through your original note where you state these things are coming later - ignore the above!

R/individuals.R
```diff
@@ -0,0 +1,66 @@
+# Define population sizes
+human_population <- 100 * 1000
```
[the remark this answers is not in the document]
giovannic: This is a great question. I've moved these population sizes into the model parameters to save headaches in the future :)

R/individuals.R
giovannic: Yes! I made the "I" (or Infected) the liverstage. It's probably worth me adding documentation to describe what these states and parameters mean. I've made an issue for that [here](https://github.com/mrc-ide/malariasimulation/issues/4)

giovannic: Prophylaxis is a great find. It looks like the later paper models the susceptability directly on humans in the "T" state. I've added a "P" state because I think it could be clearer.

PR #2, opened 2020-01-29
Feat/infection

giovannic:
This PR adds infection processes for humans and mosquitos.

One area of note is the use of one of the parameter values `dmin`. The documentation around this parameter is a bit vague, so I'm not 100% sure it's used correctly I define it [here ](https://github.com/mrc-ide/malariasimulation/pull/2/files#diff-04868f7d942b77a5834be10d0ff9427bR62) and use it [here](https://github.com/mrc-ide/malariasimulation/pull/2/files#diff-83feb0cef1c15e305d622fedd0b4f93bR309)

I also think there was a typo in one of the papers. I corrected it to make the hill function's range between 0 and 1. But it would be good to confirm that [this is correct](https://github.com/mrc-ide/malariasimulation/pull/2/files#diff-83feb0cef1c15e305d622fedd0b4f93bR303).

Things that are included in this PR:

 * a lot of new variables for humans
 * an infection_process that calculates human infections and schedules infections in the future
 * an infection scheduler that changes human states after a time delay
 * a mosquito_infection process that changes mosquitos from Susceptable to Infected
 * a treatment process which resets a human's "is_severe" variable

This PR should be reviewed after [this](https://github.com/mrc-ide/malariasimulation/pull/9) and [this one](https://github.com/mrc-ide/malariasimulation/pull/8). Since this one includes those changes


giovannic:
Obsolete

Issue #12, opened 2020-06-01
Implement Tx


giovannic:
Add treatment to the model (ACTs)

 - [x] model T and Prophylaxis states for humans
 - [x] compare to equilibrium solution

giovannic:
There are multiple drugs modelled at the same time

PR #30, opened 2020-07-17
Feat/treatment

giovannic:
This PR adds clinical treatments. (MDAs are coming soon)

This includes:
 * Parameterisation of drugs (efficacy, rel_c and prophylaxis)
 * Parameterisation of clinical treatment (drugs used and coverage)
 * A new human state `Tr` for humans undergoing treatment
 * New variables `drug`, `drug_time` and `infectivity`. To track treated humans and their onward infectivity
 * Processes and tests for the new disease progression dynamics

Notes:
 @lucyokell 
  - You may be interested in the [Treatments vignette](https://mrc-ide.github.io/malariasimulation/articles/Treatment.html), which gives an example on parameterising drugs and the effect of `ft` on prevalence.
  - Prophylaxis and onward infectivity is modelled in the `calculate_treated` function.

Please shout if you have any questions :)

@richfitz 
 - A lot of this diff is the refactor of `create_infection_process` into different functions to deal with infection, clinical infection, severe infection, treatment and scheduling.
 - I'm happy to hear any other coding suggestions.

Contingent on [this PR](https://github.com/mrc-ide/individual/pull/16). Tests will fail until I have merged and tagged that version.
Closes [this](https://github.com/mrc-ide/malariasimulation/issues/12)


giovannic:
Sorry, I had a nightmare with CI.

`devtools::test()` does not run in the same way as `CMD check`. A weird environment issue was causing my mocks to fail.

In debugging it I set up a dockerised Travis build. Should all pass now.

R/disease_progression.R
```diff
@@ -0,0 +1,64 @@
...
+ human,
+ from_state,
+ to_state,
+ rate,
+ infectivity,
+ new_infectivity
+ ) {
+  function(api) {
+    source_humans <- api$get_state(human, from_state)
+    to_move <- source_humans[bernoulli(length(source_humans), rate)]
```
richfitz, line 21: Looking at this, could you use a partial shuffle here more efficiently? As you have it `bernoulli` will do one draw per individual, but you'd draw on the order of `nh * rate` numbers with:

```
nh <- length(source_humans)
sample.int(nh, rbinom(1, nh, rate))
```

not sure if this is an important time sink of course. Similar comments apply wherever else you use bernoulli in order to identify individuals to move

R/drug_parameters.R
```diff
@@ -0,0 +1,49 @@
...
+#'
+#' @param parameters the model parameters
+#' @param ft probability of seeking treatment
+#' @param drugs vector of drugs indecies that are available
+#' @param coverages vector of coverages for those drugs
+#' @export
+set_clinical_treatment <- function(parameters, ft, drugs, coverages) {
+  parameters$ft <- ft
+  if (any(drugs < 1 | drugs > length(parameters$drug_efficacy))) {
+    stop('Drug indecies are invalid')
```
richfitz, line 41:
```suggestion
    stop('Drug indices are invalid')
```

R/drug_parameters.R
```diff
@@ -0,0 +1,49 @@
...
+  }
+  parameters$clinical_treatment_drugs <- drugs
+  if (sum(coverages) != 1) {
```
richfitz, line 44: should this be `abs(sum(coverages) - 1) < eps` (for eps ~ `sqrt(.Machine$double.eps)`) to avoid exact floating point comparisons?

R/globals.R
```diff
@@ -1 +1,7 @@
-utils::globalVariables(c("rexp", "rnorm", "runif", "stats"))
+utils::globalVariables(c(
```
richfitz, line 1: These should be imported into your namespace

```
#' @importFrom stats rexp
```

which will add a line into the `NAMESPACE` file on roxygenisation

R/individuals.R
```diff
@@ -190,6 +172,23 @@ create_variables <- function(parameters) {
     }
   )
 
+  # Initialise infectiousness of humans -> mosquitoes
+  # NOTE: not yet supporting initialisation of infectiousness of Treated individuals
+  infectivity_values <- rep(0, parameters$human_population)
+  counts <- calculate_initial_counts(parameters)
+  infectivity_values[counts[[1]]:sum(counts[1:2])] <- parameters$cd
```
richfitz, line 179: Can you document the 1:2 and 1:3 here please?

R/infection_processes.R
```diff
@@ -21,161 +21,282 @@ create_infection_process <- function(
...
+  states,
+  variables,
+  clinical_infections
+  ) {
+  parameters <- api$get_parameters()
+  seek_treatment <- bernoulli(length(clinical_infections), parameters$ft)
+  n_treat <- sum(seek_treatment)
+
+  drugs <- sample(
+    parameters$clinical_treatment_drugs,
```
richfitz, line 234: What type does `parameters$clinical_treatment_drugs` have? If numeric, you need to be careful of a terrible design decision that `sample` has...

R/parameters.R
```diff
@@ -240,6 +263,14 @@ get_parameters <- function(overrides = list()) {
     init_foim= 0,
     variety_proportions = c(.5, .3, .2),
     blood_meal_rates    = c(.92, .74, .94),
+    # treatment
+    drug_efficacy          = c(),
```
richfitz, line 267: Tiny thing, but `c()` is the same as `NULL`, not `numeric(0)`

R/parameters.R
```diff
@@ -267,7 +298,8 @@ get_parameters <- function(overrides = list()) {
     parameters$s_proportion,
     parameters$d_proportion,
     parameters$a_proportion,
-    parameters$u_proportion
+    parameters$u_proportion,
+    parameters$t_proportion
   )
 
   if (!all.equal(sum(props), 1)) {
```
richfitz, line 305: this looks like the epsilon-checking sum as mentioned in a comment above...

R/disease_progression.R
giovannic: Oh this is fantastic, thanks! It _is_ a timesink

R/infection_processes.R
giovannic: `clinical_treatment_drugs` is numeric.

Ahhh If it's of length 1 then it will sample between `1:clinical_treatment_drugs`. Oooook, I can test and deal with that. Thanks!

richfitz: the other way is to do

```
x[sample.int(x, ...)]
```

(see ?sample) which won't have that surprise in it, and is often a bit clearer than an `if/else`

PR #62, opened 2020-10-14
Feat/vector equilibrium

### LINKED ISSUE #53: parameterise_mosquito_equilibrium should figure out mosquito_limit for you
`parameterise_mosquito_equilibrium` currently asks for a `limit_grace` parameter. 

```r
parameters <- malariasimulation::get_parameters()
parameters <- malariasimulation::parameterise_mosquito_equilibrium(parameters, EIR=50, limit_grace=10)
# we write limit_grace=10 to allocate memory for 10 * total_M mosquitoes in the simulation
```

Researchers shouldn't have to worry about `limit_grace`, `mosquito_limit` or any memory allocation issues. We can use the seasonality parameters to figure that out

giovannic:
Closes #53 

Changes and tests explained in the issue

### DIFF
--- a/R/parameters.R
+++ b/R/parameters.R
@@ -416,12 +416,35 @@ parameterise_human_equilibrium <- function(parameters, eq) {
 #'
 #' @param parameters the parameters to modify
 #' @param EIR to work from
-#' @param limit_grace the number of mosquitos to allocate to the simulation as a
-#' proportion of total_M. e.g. 10 will allocate total_M * 10 mosquitos. High
-#' values of `limit_grace` are required for simulations with seasonality.
 #' @export
-parameterise_mosquito_equilibrium <- function(parameters, EIR, limit_grace=1.5) {
+parameterise_mosquito_equilibrium <- function(parameters, EIR) {
   parameters$total_M <- equilibrium_total_M(parameters, EIR)
-  parameters$mosquito_limit <- parameters$total_M * limit_grace
+  K0 <- calculate_carrying_capacity(parameters)
+  R_bar <- calculate_R_bar(parameters)
+  max_K <- max(vnapply(seq(365), function(t) {
+    carrying_capacity(
+      t,
+      parameters$model_seasonality,
+      parameters$days_per_timestep,
+      parameters$g0,
+      parameters$g,
+      parameters$h,
+      K0,
+      R_bar
+    )
+  }))
+  omega <- calculate_omega(parameters)
+  max_total_M <- max_K * (
+    1 / (
+      2 * parameters$dl * parameters$mum * (1 + parameters$dpl * parameters$mup)
+    )
+  ) * (
+    1 / (
+      parameters$gamma * (omega + 1)
+    )
+  ) * (
+    omega / (parameters$ml * parameters$del) - 1 / (parameters$ml * parameters$dl) - 1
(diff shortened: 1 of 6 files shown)


tests/testthat/test-biology.R
```diff
@@ -82,3 +82,35 @@ test_that('total_M and EIR functions are consistent with equilibrium EIR (with h
     tolerance = 1
   )
 })
+
+test_that('mosquito_limit is set to a sensible level', {
+  population <- 1000
+
+  EIRs <- c(5, 50, 1000)
```
pwinskill, line 89: worth having somthing like 
`expect_equal(equilibrium_total_M(parameters, 0), 0` 
here?

src/mosquito_emergence.cpp
```diff
@@ -36,7 +37,11 @@ Rcpp::XPtr<process_t> create_mosquito_emergence_process_cpp(
             }
             auto source = api.get_state(mosquito, unborn);
             if (source.size() < n) {
-                Rcpp::stop("Not enough mosquitos. Please raise mosquito_limit");
+                std::stringstream m;
+                m << "Not enough mosquitos (short by ";
+                m << n - source.size();
```
pwinskill, line 42: Not sure it would be obvious to the user where to go to fix this issue if it occured. Maybe the message could point them towards  `parameters$mosquito_limit`,  and/or give some indication that checking the input seasonality (and EIR?) might be worthwhile?

R/parameters.R
```diff
@@ -414,14 +414,43 @@ parameterise_human_equilibrium <- function(parameters, eq) {
...
+      )
+    )
+  ) * (
+    1 / (
+      parameters$gamma * (omega + 1)
+    )
+  ) * (
+    omega / (parameters$ml * parameters$del) - (1 / (parameters$ml * parameters$dl)) - 1
+  )
+  parameters$mosquito_limit <- ceiling(max_total_M * 5) #Allow for random fluctuations
```
pwinskill, line 454: Is there some top limit here that shoudl throw an error? Maybe its checked elsewhere but if I put in a combination of wrong seasonality pars and and very high EIR will the run just choke my computer memory?

pwinskill (commented):
Hey Giovanni,
Looks good - hopefully just removes one more ask from the user. I guess there could be edge cases where the automatic approach fails, but if I've understood correctly the user could override the parameter setting if appropriate to do so.

R/parameters.R
giovannic: That's a good question!

Seasonality and EIR combinations feel like a complicated parameter set to validate. We'd need to come up with pretty robust rules to only throw errors when the researcher is genuinely making a mistake and not plotting extremes or experimenting. My guess would be that we don't have those rules?

On large scale execution, the model will crash, get an R error, or be killed by the OS before it consumes all memory. I've just run a bunch of tests to confirm that.

On that note, we should make an issue for a maximum human/EIR to support. Let's say `human_population=1e6`, `EIR=1e5`?

pwinskill: Agreed - I don't think we have those rules so as long as it fails (somewhat) elegantly I think thats fine!
Yes, those seem like very generous top limits  for pop and EIR

pwinskill (approved):

PR #80, opened 2021-03-09
Simplifications and biology fixes:

giovannic:
* Create a mosquito death event
* Add validation for species parameterisation
* Translate ODE stepper process to R
* Fix carrying capacity bug
* Make stochastic Extrinsic Incubation Period


R/events.R
```diff
@@ -166,6 +168,14 @@ attach_event_listeners <- function(
     individual::update_category_listener(variables$mosquito_state, 'Im')
   )
 
+  events$mosquito_death$add_listener(
+    function(timestep, target) {
+      variables$mosquito_state$queue_update('Unborn', target)
```
pwinskill, line 173: Minor point - but (speaking with my entomologist hat on) its not clear to me what "Unborn" as a mosquito state refers to. It could mean an individual that doesn't exist, an egg, a larvae. a pupae... Perhaps we can make it a bit more clear?

R/vector_parameters.R
```diff
@@ -53,6 +53,12 @@ set_species <- function(parameters, species, proportions) {
   if (length(species) != length(proportions)) {
     stop('You must give proportions for each species')
   }
+  if (any(proportions <= 0)) {
```
pwinskill, line 56: Shouldn't this be if `(any(proportions < 0))`
We can specify 0 proportions right?

giovannic:
Thanks, I've:

 * added fixes and tests to make sure the ODE survives with total_M = 0
 * renamed Unborn to NonExistent

pwinskill (approved):
👍

PR #81, opened 2021-03-11
Feat/pois bites

giovannic:
To review after #80 

A more correct biting model:

 * bites are modelled per species
 * number of bites are modelled using a poisson process
 * the bitten humans are sampled using their effective biting rate per
   individual (taking into account heterogeneity, age and interventions)


pwinskill (commented):
Hey!
Still getting used to the bitset operations, so a nice opportunity to see some more complex use cases. The new biting process stuff all seems to make sense to me.

PR #83, opened 2021-03-16
Set time varying treatments

giovannic:


R/human_infection.R
```diff
@@ -149,20 +149,26 @@ calculate_treated <- function(
   parameters,
   timestep
   ) {
-  if (length(parameters$clinical_treatment_coverages) == 0) {
+  treatment_coverages <- get_treatment_coverages(parameters, timestep)
+  ft <- sum(treatment_coverages)
+  if (ft > 1) {
+    stop('Drug coverages need to be < 1 at all timesteps')
```
pwinskill, line 155: I think this can be clarified, perhaps:
`stop('Drug coverages must sum to <= 1 at all timesteps')`
?

pwinskill: No idea if this is the case, but would it be more efficiecnt to have this sum <=1 check run once in R/drug_parameters.R?

pwinskill (commented):
Looks great 😎

PR #89, opened 2021-04-08
Feat/delay

giovannic:
Add sophisticated delays to the adult dde model.

 * Create a History class to manage storage, interpolation and validation
 * Integrate History into the differential equations for the adult mosquito model
 * Update individual to interface with the new solver


src/history.cpp
```diff
@@ -0,0 +1,49 @@
...
+    while(it->first > time) {
+        // Check if we're at the start of the history
+        if (it == values.begin()) {
+            Rcpp::stop("`time` is before our stored history");
+        }
+        it--;
+    }
+
+    // Interpolate
+    return it->second + (
```
richfitz, line 46: It looks like here you are doing a linear interpolation of a variable. Depending on the dynamics of the system this may cause your error calculation to lock up because the linear error here (which will echo through the solution) will be impossible for it to eliminate even as it reduces the step size, as the error is already baked in. I am not sure in practice how much of a problem this is, especially if the mosquito system is simple - it's a bit like having a forcing function that is piecewise linear, which we include all the time.

This is the interface for "dense output" which allows interpolation within a timestep with the odint solvers:
https://www.boost.org/doc/libs/1_60_0/libs/numeric/odeint/doc/html/boost_numeric_odeint/concepts/dense_output_stepper.html

However, it's not at all clear to me how you'd be able to push back the that state each step.

src/history.h
```diff
@@ -0,0 +1,25 @@
...
+ * history.h
+ *
+ *  Created on: 08 Apr 2021
+ *      Author: gc1610
+ */
+
+#ifndef SRC_HISTORY_H_
+#define SRC_HISTORY_H_
+
+#include <map>
```
richfitz, line 11: nice idea :)

src/solver.h
```diff
@@ -37,6 +43,7 @@ struct Solver {
     const double dt = 1.;
     state_t state;
     integration_function_t ode;
+    observer_t observer;
```
richfitz, line 46: You might need to add an additional control parameter to set a max step size equal to the delay length. Otherwise when you go back to find the interval that bounds `t - tau` you won't find it because it's part of the current step when `h > tau`

richfitz (changes requested):

src/history.cpp
giovannic: I think this is important, so I'm gonna delay 😉 this fix until I get some time to figure out the most efficient way to do this.

My hope is that I can keep a history in boost's stepper and use `calc_state`. At the moment, I don't think the boost's stepper has the history.

giovannic:
Closing until this becomes a priority

PR #93, opened 2021-04-21
Bug/competing hazards

giovannic:
There there was a competition between:

 * force of infection (S,U,A -> D) and disease progressions (A -> U and U -> S)
 * force of asymptomatic infection (S,U -> A) and disease progressions (U -> S)

details commented in the code

Changes:

 * Separate asymptomatic infection from progression events for easy monitoring
 * Draw disease progression every timestep to avoid it being pre-empted by infection
 * Render outputs to monitor transition rates


R/disease_progression.R
```diff
@@ -19,30 +19,19 @@ create_infection_update_listener <- function(
...
-#' @title Modelling the progression of the human disease
-#' @description schedules follow up infection events with the log uniform dist.
-#'
-#' @param event the event to schedule
-#' @param rate the average time spent in this state
-#' @noRd
-create_progression_listener <- function(event, rate) {
+create_rate_listener <- function(from_state, to_state, renderer) {
   function(timestep, target) {
-    event$schedule(target, log_uniform(target$size(), rate))
```
giovannic, line 45: Disease progression (D -> A -> U -> S) was scheduled in advance.

As soon as an individual arrived in U we would lock in when they transitioned to S.

Note: if any other process calls event$clear_schedule(...) the rate will be reduced

R/disease_progression.R
```diff
@@ -19,30 +19,19 @@ create_infection_update_listener <- function(
...
-#' @noRd
-initialise_progression <- function(event, state, from_state, rate) {
-  target <- state$get_index_of(from_state)
-  event$schedule(
-    target,
-    log_uniform(target$size(), rate)
-  )
+create_progression_process <- function(event, state, from_state, rate) {
+  function(timestep) {
+    event$schedule(state$get_index_of(from_state)$sample(1/rate), 0)
```
giovannic, line 24: Now disease progression (D -> A -> U -> S) is scheduled in a memoryless way.

Every timestep, we draw an individual from U to transition to S.

Note: if an infection is scheduled for the same timestep, the event order will resolve what happens (more below)

R/disease_progression.R
```diff
@@ -19,30 +19,19 @@ create_infection_update_listener <- function(
...
-#' @noRd
-create_progression_listener <- function(event, rate) {
+create_rate_listener <- function(from_state, to_state, renderer) {
   function(timestep, target) {
-    event$schedule(target, log_uniform(target$size(), rate))
+    renderer$render(
+      paste0('rate_', from_state, '_', to_state),
+      target$size(),
+      timestep
+    )
```
giovannic, line 34: The red flag was that the output of rate_U_S / U_count, was much lower than the progression rate we parameterised 1/du.

R/events.R
```diff
@@ -1,12 +1,16 @@
 create_events <- function(parameters) {
   events <- list(
+    # Disease progression events
```
giovannic, line 3: We re-order events to deal with the case where individuals are scheduled to progress to the next stage of disease AND be re-infected.

asymptomatic_progression (D -> A), subpatent_progression (A -> U) and recovery (U -> S) are now earlier in the event list than clinical_infection (S, U, A) -> D and asymptomatic_infection (S, U -> A).

So If an individual is due to progress from U -> S AND be infected to D at the same timestep. The clinical_infection updates will overwrite the subpatent_progression updates.

R/human_infection.R
```diff
@@ -309,8 +309,6 @@ schedule_infections <- function(
       infection_times
     )
     events$detection$schedule(to_infect, infection_times)
-    events$subpatent_infection$clear_schedule(to_infect)
```
giovannic, line 312: On infection, we would cancel infected individuals progressing to their next state.

Initially this makes sense. If you are infecting from U -> D, the individual is already scheduled to go to S and if you don't cancel it, your infection won't last!

But this is the cause of the competing hazard. By cancelling subpatent_infection and recovery (A -> U -> S) we reduce the rates `da` and `du`.

Instead, for recovery, we want to sample individuals from U who are NOT going to be reinfected.

R/disease_progression.R
```diff
@@ -19,30 +19,31 @@ create_infection_update_listener <- function(
...
+  rate
+  ) {
+  function(timestep) {
+    incubating <- events$clinical_infection$get_scheduled()$or(
+      events$asymptomatic_infection$get_scheduled()
+    )
+    progression_event$schedule(
+      state$get_index_of(from_state)$and(incubating$not())$sample(1/rate),
+      0
+    )
```
giovannic, line 36: Now disease progression (D -> A -> U -> S) is scheduled in a memoryless way.

Every timestep, we draw an individual from U to transition to S.

We exclude individuals who are incubating. However, one could imagine a model where individuals continue to recover until they are reinfected when their incubation period is over.

R/events.R
pwinskill: I'm wondering if this still does not capture the competing hazards completely?
If I've understood correctly this is saying that the hazard of infection dominates the "hazard" of  recovery (U->S)?
I don't know how it was dealt with previously, but I was thinking of the following:
```
# Probability of infection
prob_infection = 0.25
# Probability of recovery
prob_recover = 0.5
# Probability of infection OR recovery 
prob_either = 1 - (1 - prob_infection) * (1 - prob_recover)
# Draw if either happens to the individual
runif(1) < prob_either
# If either does happen, sample which of the two outcomes occurs:
sample.int(2, c(prob_infection, prob_recover))
```

R/biting_process.R
```diff
@@ -18,30 +18,78 @@ create_biting_process <- function(
...
+    events$subpatent_progression$schedule(which(outcomes == 2), 0)
+    events$recovery$schedule(which(outcomes == 3), 0)
+  }
+}
+
+#' @importFrom stats runif 
+simulate_competing_outcomes <- function(prob_outcome) {
+  prob_occurs <- apply(prob_outcome, 1, function(p) {
+    1 - prod(1 - p)
+  })
```
pwinskill, line 75: This apply is painfully inefficient, maybe add this little bit to cpp:
```

f1 <- function(prob_outcome){
  prob_occurs <- apply(prob_outcome, 1, function(p) {
    1 - prod(1 - p)
  })
  return(prob_occurs)
}

library(Rcpp)

cppFunction('NumericVector f2(NumericMatrix x) {
  int nrow = x.nrow();
  int ncol = x.ncol();
  NumericVector out(nrow);

  for (int i = 0; i < nrow; i++) {
    double row_product = 1;
    for (int j = 0; j < ncol; j++) {
      row_product *= 1 - x(i,j);
    }
    out[i] = 1 - row_product;
  }
  return out;
}')

test_matrix <- matrix(runif(10000 * 3), ncol = 3)


o1 <- f1(test_matrix)
o2 <- f2(test_matrix)
all.equal(o1, o2)

microbenchmark::microbenchmark(f1(test_matrix), f2(test_matrix))
```

Gives:
```
Unit: microseconds
            expr     min       lq      mean   median       uq     max neval
 f1(test_matrix) 12747.7 14051.25 16822.887 15002.45 17654.85 76961.7   100
 f2(test_matrix)    23.8    26.70    46.031    36.10    40.95   903.9   100
```

R/biting_process.R
```diff
@@ -18,30 +18,78 @@ create_biting_process <- function(
...
+  n_occurances <- sum(occurs)
+  if (n_occurances == 0) {
+    return(outcomes)
+  }
+
+  if (n_occurances == 1) {
+    # apply doesn't work for vectors in the same way
+    p <- prob_outcome[occurs, ]
+    outcomes[occurs] <- sample.int(n = length(p), size = 1, prob = p)
+  } else {
```
pwinskill, line 87: You can force this to keep the output as a matrix, even if `n_occurances = 1`, by adding `,drop = FALSE`
Example:
```
x <- matrix(1:4, ncol = 2)
# Turns to vector
x[c(TRUE, FALSE),]
# Keep as matrix
x[c(TRUE, FALSE), ,drop = FALSE]
```
then I think you wouldn't need this additional if statement

pwinskill (commented):
Looks good - will be interesting to see if we observe much impact of these changes, although I guess it might help with some cases where we saw differences vary alot with respect to the EIR

R/biting_process.R
giovannic: That's amazing! Tested that and put it in. 5x speedup

PR #108, opened 2021-05-20
Update vector_parameters.R

nilani-chandradeva1:
added some potential koliensis data

### DIFF
--- a/R/vector_parameters.R
+++ b/R/vector_parameters.R
@@ -46,6 +46,20 @@ fun_params <- list(
   mum = .112
 )
 
+kol_params <- list(
+  species = 'kol',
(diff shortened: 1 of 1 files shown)


R/vector_parameters.R
```diff
@@ -46,6 +46,20 @@ fun_params <- list(
   mum = .112
 )
 
+kol_params <- list(
```
giovannic, line 49: You've missed out the `@export` tag. Here's some [documentation on it](https://kbroman.org/pkg_primer/pages/docs.html)
```suggestion
#' @export
kol_params <- list(
```

giovannic (changes requested):
We need a bit more documentation for this

PR #112, opened 2021-05-21
Fix tbv bug

### LINKED ISSUE #109: TBV not affecting prevalence or incidence
Hi Giovanni,
I've run the vaccines vignette, implementation of the transmission-blocking vaccine isn't affecting prevalence or incidence of malaria in the population (easier to ascertain by increasing population size or turning off seasonality). The RTS,S vaccine seems fine. I'm not quite sure what's causing the issue, but the version of set_tbv() we were using last year contained the arguments 'start', 'end' & 'frequency', instead of 'timesteps'. So I wondered whether this is causing the issue. Happy to discuss in more detail if useful, and there's no huge rush to sort out. Thanks!

giovannic:
So it turns out that the FOIM was not updating from the equilibrium value.

Closes #109 

=== longer explanation ===

We lag FOIM by delay_gam (12.5 timesteps) and use LaggedValue objects to control that. They have a fixed buffer size, which we set to 13 timesteps. Every timestep we saved a new FOIM _before_ getting the FOIM from 12.5 timesteps ago. So the LaggedValue object was deleting the oldest FOIM value and falling back on the initial FOIM for the simulation

====================

Here are some graphs with new Vaccine vignette outputs...

tbv_infectivity
foim_1
tbv

### DIFF
--- a/R/events.R
+++ b/R/events.R
@@ -228,31 +228,7 @@ attach_event_listeners <- function(
 
   if (parameters$tbv == 1) {
     events$tbv_vaccination$add_listener(
-      function(timestep) {
-        time_index = which(parameters$tbv_timesteps == timestep)
-        target <- which(trunc(get_age(
-          variables$birth$get_values(),
-          timestep
-        ) / 365) %in% parameters$tbv_ages)
-        to_vaccinate <- which(sample_intervention(
-          target,
-          'tbv',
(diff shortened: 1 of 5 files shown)


JDChallenger (approved):
Hi Giovanni, this looks good, thanks! Joe

PR #115, opened 2021-06-03
Time varying vector controls

### LINKED ISSUE #58: Vector control resistance
Currently bednet paramters `rn` and `dn` are are constant for the whole of the simulation:

```r
# ...
# set bednet profile
bednetparams <- get_parameters(list(
    rn = .56,
    rnm = .24,
    dn0 = .533,
    gamman = 2.64 * 365
))

# set bednet strategy
bednetparams <- set_bednets(
  bednetparams,
  timesteps = 4:7 * year,
  coverages = rep(.8, 4),
  retention = 5 * year
)
# ...
```

But in actuality mosquitoes become resistant to bednets. We would like to model changes to dn/rn. Questions:

 * are the dn/rn changes normally modelled as a curve or discrete changes?
 * And are these changes independent of the type of bednet? See #54

giovannic:
Closes #58 

 * New IRS model
 * Update ITN model for time-varying dn0 and rn
 * Vignette for time varying vector control parameters

Example vignette run...
[Vector Control.pdf](https://github.com/mrc-ide/malariasimulation/files/6646998/Vector.Control.pdf)

### DIFF
--- a/R/biting_process.R
+++ b/R/biting_process.R
@@ -8,6 +8,8 @@
 #' @param variables a list of all of the model variables
 #' @param events a list of all of the model events
 #' @param parameters model pararmeters
+#' @param lagged_infectivity a LaggedValue class with historical sums of infectivity
+#' @param lagged_eir a LaggedValue class with historical EIRs
 #' @noRd
 create_biting_process <- function(
   renderer,
@@ -16,7 +18,7 @@ create_biting_process <- function(
   variables,
   events,
   parameters,
-  lagged_foim,
+  lagged_infectivity,
   lagged_eir
   ) {
   function(timestep) {
@@ -32,7 +34,7 @@ create_biting_process <- function(
       age,
       parameters,
       timestep,
-      lagged_foim,
+      lagged_infectivity,
       lagged_eir
     )
 
@@ -58,7 +60,7 @@ simulate_bites <- function(
   age,
   parameters,
   timestep,
-  lagged_foim,
+  lagged_infectivity,
   lagged_eir
   ) {
   bitten_humans <- individual::Bitset$new(parameters$human_population)
@@ -95,30 +97,8 @@ simulate_bites <- function(
     W <- average_p_successful(p_bitten$prob_bitten_survives, .pi, Q0)
     Z <- average_p_repelled(p_bitten$prob_repelled, .pi, Q0)
     f <- blood_meal_rate(s_i, Z, parameters)
-
-    lambda <- .effective_biting_rates(
-      s_i,
-      psi,
-      .pi,
-      zeta,
-      p_bitten,
-      W,
-      Z,
-      f,
-      parameters
-    )
-
-    renderer$render(
-      paste0('lambda_', parameters$species[[s_i]]),
-      sum(lambda),
-      timestep
-    )
-
-    renderer$render(
-      paste0('normal_lambda_', parameters$species[[s_i]]),
-      sum(lambda) / mean(psi),
-      timestep
-    )
+    a <- .human_blood_meal_rate(f, s_i, W, parameters)
+    lambda <- effective_biting_rates(a, .pi, p_bitten)
 
     if (parameters$individual_mosquitoes) {
       species_index <- variables$species$get_index_of(
@@ -136,18 +116,19 @@ simulate_bites <- function(
       n_infectious <- calculate_infectious_compartmental(solver_states)
     }
 
-    lagged_eir[[s_i]]$save(n_infectious * sum(lambda), timestep)
+    lagged_eir[[s_i]]$save(n_infectious * a, timestep)
     species_eir <- lagged_eir[[s_i]]$get(timestep - parameters$de)
-    EIR <- EIR + species_eir / sum(psi)
-    n_bites <- rpois(1, species_eir)
+    EIR <- EIR + species_eir
+    n_bites <- rpois(1, species_eir * mean(psi))
     if (n_bites > 0) {
       bitten_humans$insert(
         fast_weighted_sample(n_bites, lambda)
       )
     }
 
-    foim <- calculate_foim(human_infectivity, lambda, psi)
-    lagged_foim$save(foim, timestep)
+    infectivity <- lagged_infectivity$get(timestep - parameters$delay_gam)
+    lagged_infectivity$save(sum(human_infectivity * .pi), timestep)
+    foim <- calculate_foim(a, infectivity)
     renderer$render(paste0('FOIM_', s_i), foim, timestep)
     mu <- death_rate(f, W, Z, s_i, parameters)
     renderer$render(paste0('mu_', s_i), mu, timestep)
@@ -161,7 +142,7 @@ simulate_bites <- function(
 
       biting_effects_individual(
         variables,
-        lagged_foim$get(timestep - parameters$delay_gam),
+        foim,
         events,
         s_i,
         susceptible_species_index,
@@ -174,7 +155,7 @@ simulate_bites <- function(
       adult_mosquito_model_update(
         models[[s_i]],
         mu,
-        lagged_foim$get(timestep - parameters$delay_gam),
+        foim,
         solver_states[[ADULT_ODE_INDICES['Sm']]],
         f
       )
@@ -192,37 +173,15 @@ simulate_bites <- function(
 # =================
 
 calculate_eir <- function(species, solvers, variables, parameters, timestep) {
-  lambda <- effective_biting_rates(species, variables, parameters, timestep)
+  a <- human_blood_meal_rate(species, variables, parameters, timestep)
   infectious <- calculate_infectious(species, solvers, variables, parameters)
-  infectious * sum(lambda)
-}
-
-effective_biting_rates <- function(species, variables, parameters, timestep) {
   age <- get_age(variables$birth$get_values(), timestep)
   psi <- unique_biting_rate(age, parameters)
-  zeta <- variables$zeta$get_values()
-  p_bitten <- prob_bitten(timestep, variables, species, parameters)
-  .pi <- human_pi(zeta, psi)
-  Q0 <- parameters$Q0[[species]]
-  W <- average_p_successful(p_bitten$prob_bitten_survives, .pi, Q0)
-  Z <- average_p_repelled(p_bitten$prob_repelled, .pi, Q0)
-  f <- blood_meal_rate(species, Z, parameters)
-  .effective_biting_rates(species, psi, .pi, zeta, p_bitten, W, Z, f, parameters)
+  infectious * a * mean(psi)
 }
 
-.effective_biting_rates <- function(
-  species,
-  psi,
-  .pi,
-  zeta,
-  p_bitten,
-  W,
-  Z,
-  f,
-  parameters
-  ) {
-  a <- human_blood_meal_rate(f, species, W, parameters)
-  a * intervention_coefficient(p_bitten) * zeta * psi
+effective_biting_rates <- function(a, .pi, p_bitten) {
+  a * .pi * p_bitten$prob_bitten / sum(.pi * p_bitten$prob_bitten_survives)
 }
 
 calculate_infectious <- function(species, solvers, variables, parameters) {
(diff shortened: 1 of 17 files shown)


EllieSherrardSmith:
There is one change to make for the spraying_decay function for blood feeding (ks) 

We have to use two functions I think, the first is fine (in R/vector_control.R LINES 186-188)
then we can add:

spraying_decay_fed <- function(t, theta, gamma) {​

k0 / (1 + exp(-(theta + gamma * t)))
}

And change line 54 (ks <- spraying_decay(since_spray, ks_theta, ks_gamma)) to
ks <- spraying_decay_fed(since_spray, ks_theta, ks_gamma, k0) 
or may need to be
ks <- spraying_decay(since_spray, ks_theta, ks_gamma, parameters
)

the reason is that the data indicate that the successful feeding can be greater that the value of k0 (which is an assumed parameter for the feeding in the absence of any intervention). So to keep this between 0 and 1 we fix the numerator to k0.

EllieSherrardSmith (changes requested):
In **R/vector_control.R**

**please change line 54** from 
ks <- spraying_decay(since_spray, ks_theta, ks_gamma)

to 
ks <- spraying_decay_fed(since_spray, ks_theta, ks_gamma, k0)
or
ks <- spraying_decay_fed(since_spray, ks_theta, ks_gamma, parameters)
where parameters$k0 is set at 0.699

And, **please add (after line 188)**:

spraying_decay_fed <- function(t, theta, gamma, k0) {
  k0 / (1 + exp(-(theta + gamma * t)))
}

or

spraying_decay_fed <- function(t, theta, gamma, parameters) {
  k0 / (1 + exp(-(theta + gamma * t)))
}


In **man/arab_params.Rd**
In **man/fun_params.Rd**
In **man/gamb_params.Rd**

can we use the parameters as defined in https://www.pnas.org/content/pnas/early/2019/07/02/1820646116.full.pdf

Which are: 
phi_bednets: 0.80 [[arab]]
phi_indoors: 0.86 [[arab]]

phi_bednets: 0.78 [[fun]]
phi_indoors: 0.87 [[fun]]

phi_bednets: 0.85 [[gamb]]
phi_indoors: 0.90 [[gamb]]


In **man/get_parameters.Rd**

Same applies - perhaps use line 166
\item phi_bednets - proportion of bites taken in bed; default = 0.85

and line 169
\item phi_indoors - proportion of bites taken indoors; default = 0.90


In **man/set_spraying.Rd**

Could you annotate this as per nets to confirm we are inputting matrix with dimensions (length(species) * length(time_steps))

EllieSherrardSmith (changes requested):
Hello 

I think this **R/vector_control.R** is not updated?

Line 187 is not correct currently for line 54
1 / (1 + exp(-(theta + gamma * t)))
the function needs to be:
k0 / (1 + exp(-(theta + gamma * t)))
just for the ks estimate (line 54)

this is because we otherwise get estimates above 1 or below 0 as k0 is 0.699
though I might be mis-understanding the math if this is included at a different part of the eqns - in which case could you explain :)

EllieSherrardSmith (commented):
Hi Giovanni 

I think there is an error for the IRS (it looks a little different in the outputs from Jamie's irs_restructure2 branch which is what I used for the RCTs)

Could you take a look at 

https://github.com/mrc-ide/Malaria_model/blob/irs_restructure2/code/mosq.cpp 
**Line 446**

I think in the **R/vector_control.R** code we need to change 
**line 162-163** prob_suvives_spraying
to 
(1 - ks_prime/irs_k0)*(**lp_irs**/(lp_irs + jp_irs))
instead of ks_prime/k0

and then these wrap up into w_irs (prob of biting)

w_irs=(1-r2)*(1-d2);
w_itn_irs=w_irs*(1-d1-r1);

BUT this is multiplicative if we use both nets AND sprays, so I think we need another if else in the prob_bitten_survives functon that would reassses **biting_process.R** at lines 201 and 202

I think something like

if (parameters$spraying & paramters$bednets) {}

But, I am unsure if I am correctly tracking this through to biting_process.R

The result as things stand is that the IRS looks too good relative to previous simulations. I think either the code change of  **R/vector_control.R** at **line 162-163** or the multiplicative benefit would resolve this :/ 

Thanks

giovannic:
Hey this is great! I've looked over the issues. And there may be some differences but I'm a bit hesitant to change malariasimulation to match the old structure...

### First thing: should use `d_s` instead of `s_s`?

For context, these are the equations I'm referring to:
Screenshot 2021-06-03 at 13 54 15

So `s_s = 1 - r_s - d_s`. I've managed to confirm it with some rearranging using `j_prime = 1 - l_prime - k_prime`. This also makes sense intuitively, a feeding attempt is successful if a mosquito is not repelled or killed.

But `s_s` seems to be a simpler calculation than `d_s`. So there less likely to be mistakes and rounding errors calculating this way. Can we keep it this way?

### Second thing: probability of success vs probability of failure

For context, malariasimulation recreates the probabilities from this table:

Screenshot 2021-06-25 at 15 17 13

[This code](https://github.com/mrc-ide/malariasimulation/blob/feat/resistance/R/vector_control.R#L83-L96) maps quite nicely onto the 3rd column.

But you'll notice that when spraying is FALSE, [these parameters](https://github.com/mrc-ide/malariasimulation/blob/feat/resistance/R/vector_control.R#L76-L79) are set to make them equal to the second column.

And when bednets is FALSE, [these parameters](https://github.com/mrc-ide/malariasimulation/blob/feat/resistance/R/vector_control.R#L35-L37) are set making them equal to the first column.

I've double checked those calculations but it looks alright to me. Do you agree?

At Malaria_Model->Mosq.cpp:446, calculates the probability of failure instead of success. Intuitively it makes sense, and I suspect that it's an equivalent approach (but I've not verified!). But probability of failure is undocumented in any of our published papers and has implications for the FOIM calculations.

Can we stay with probability of a successful bite?

### Third thing: Are we sure about prob_repeat?

If [Malaria_Model->mosq.cpp->prob_repeat](https://github.com/mrc-ide/Malaria_model/blob/irs_restructure2/code/mosq.cpp#L497-L500) is the probability that a mosquito is repelled (third row of the probability matrix), it looks like line 498 should instead be:

```cpp
// was previously z_itn_irs=r2+(1-r2)*r1;
z_itn_irs=(1-r2)*r1;
````

But maybe I'm misunderstanding what `prob_repeat` is about.

### Fourth thing: We have recently fixed an FOIM bug

You can find details on it [here](https://github.com/mrc-ide/malariasimulation/pull/120). I've updated this branch with the same fix to that bug.

It could have implications for total_M in your experiment and have different outputs. Could you re-install this branch and check if this changes your results?

EllieSherrardSmith:
**point 1 - 3** are linked _i think_

I have an email chain discussion with Jamie Griffin from last year where he confirmed this all (copied below) - but looking at the code something has been altered 9 months ago on the C++ irs_structure2 branch (not sure why here)... can we chat? 

I think it is to do with the assumption that death from irs would happen before the mosquito is repeating the behaviour (r_irs)
So the unconditional probability is distinct as ds and rs and ss might not have to sum to 1 ... I am a bit hazy on this... and it might be ok algebraically but I think we need to code in the ds to track what the original model did? 

then we can work out 
w_irs = (1-rs)*(1-ds) 
w_itn_irs = w_irs * (1 - dn - rn)

(just for clarity if we revisit this later r2 is synon with rs in the two codes, d2 is ds, and d1 is dn as r1 is rn... )


It may well be my inability to algebraically rearrange the maths at the moment!

Here is the email exchange:

Jamie Griffin <j.griffin@qmul.ac.uk>
Tue 10/6/2020 10:54 PM
Hi Ellie,

 We're fine, thanks.

 That part of the C++ code has been the same since I originally implemented it. This is the BM-type description I was originally sent in early 2017:

r_IRS = if TIME < IRS_ITN_on 0 else (1 – kp_IRS/k0) x (jp_IRS/(lp_IRS + jp_IRS))
d_IRS = if TIME < IRS_ITN_on 0 else (1 – kp_IRS/k0) x (lp_IRS/(lp_IRS + jp_IRS))
s_IRS = if TIME < IRS_ITN_on 1 else  kp_IRS/k0

kp_IRS = (1 – det_hut_IRS) x succ_hut_IRS
jp_IRS = (1 – det_hut_IRS) x rep_hut_IRS + det_hut_IRS
lp_IRS = (1 – det_hut_IRS) x mort_hut_IRS

det_hut_IRS = 1 / (1 + exp-(IRS_decay_DET1 + IRS_decay_DET2 x TIME (after IRS_ITN_on) ))
succ_hut_IRS = 1 / (1 + exp-(IRS_decay_SUCC1 + IRS_decay_SUCC2 x TIME (after IRS_ITN_on) ))
mort_hut_IRS = 1 / (1 + exp-(IRS_decay_MORT1 + IRS_decay_MORT2 x TIME (after IRS_ITN_on) ))

rep_hut_IRS = 1 -  succ_hut_IRS -  mort_hut_IRS

Here r_IRS + d_IRS + s_IRS = 1. So the probabilities are the unconditional probabilities for the outcomes of a feeding attempt.

This corresponds to rs, ds and ss in the R code you sent.

But in the C++ code, r2 occurs before d2, i.e. the death is conditional on not previously being repelled, so d2+r2 couldgreater than 1. So the unconditional probability of death is d2*(1-r2)=d_IRS, and d2=d_IRS/(1-r2).

Best wishes,
Jamie

From: Sherrard-Smith, Ellie <e.sherrard-smith@imperial.ac.uk>
Sent: 06 October 2020 00:25
To: Jamie Griffin <j.griffin@qmul.ac.uk>
Subject: Re: model code on Github
 
Hi Jamie,

How are you? I hope you and your family are holding up ok through the pandemic

I had a query on the coding we have adjusted for IRS. I noticed that in the mosq.cpp line 446,

the d2 (ds parameter) is calculated as I would expect but then divided by (1 - r2). I am not clear why the /(1 - r2) is present. Is this an error? Or am I mis-understanding the algebra! I had a look at this and made the attached script to explain why I think it should be
d2 = (1 - kp_s/irs_k0)*(lp_s/(lp_s + jp_s)) 

Could you confirm either way and I can then update these - thank you

Ellie

## the file i sent was this:

vector_control_decay_sprays_rs_ss <- function(since_spray, 
                                              irs_decay_mort1, irs_decay_mort2,
                                              irs_decay_succ1, irs_decay_succ2,
                                              irs_k0, # the maximum successful feeding in the absence of indoor interventions
                                              irs_decay_det1, irs_decay_det2) {
  
  mort = 1 / (1 + exp(-(irs_decay_mort1 + irs_decay_mort2 * since_spray)))
  succ = irs_k0 / (1 + exp(-(irs_decay_succ1 + irs_decay_succ2 * since_spray)))
  det =  1 / (1 + exp(-(irs_decay_det1  +  irs_decay_det2 * since_spray)))
  rep = 1 - mort - succ
  
  kp_s = (1 - det) * succ
  jp_s = (1 - det) * rep + det
  lp_s =  (1 - det) * mort
  
  rs = (1 - kp_s/irs_k0)*(jp_s/(lp_s + jp_s))
  ds = (1 - kp_s/irs_k0)*(lp_s/(lp_s + jp_s)) ## my understanding is that it should be this ... from mosq.cpp  
  # ds = (1 - kp_s/irs_k0)*(lp_s/(lp_s + jp_s))/(1-rs) ## this is from irs_restructure.exe (mosq.cpp) - not sure why we have /(1-rs) 
  # i am unsure if i am correctly interpreting this as may be algebraically correct..! 
  ss = 1 - rs - ds ## this makes sense to me and is what I thought we were doing
  
}

There is a confusion though as the code has been tweaked and I am unsure why? 
However, Jamie's published code has it how I thought it should be given the email exchange above (slightly different as also larviciding in)...
https://github.com/jamiegriffin/Malaria_simulation/blob/master/code/mosq.cpp (around lines 511)

**point 4**
Great thank you - have pulled this.

Here is where we are with the same parameters at the moment (unfortunately I cannot show last 2 data points yet but if we talk i can demonstrate those (we match nets ok i think so think this is an IRS issue.)

The decay of impact is ever-so-slightly faster in Jamie's code than in malariasimulation (am really picking tiny things here!!! but good to check) The yellow and darker blue have IRS and do a little better than I am expecting them to

check model agreement1


Great to chat if that is useful. 

Thanks

EllieSherrardSmith (approved):
This looks good to me and gives closer to matched outcomes then ever! Happy to accept these

PR #120, opened 2021-06-22
Fix FOIM normalisation bug:

giovannic:
* FOIM was not being corrected for age, as is done in the literature (https://doi.org/10.1371/journal.pmed.1000324, protocol 1, page 6)
* Human infectivity towards mosquitoes was initialised incorrectly
* Update missing documentation

The formula for FOIM is...

Screenshot 2021-06-22 at 11 53 00

Where omega is...
Screenshot 2021-06-22 at 11 53 41

Omega was left out of this implementations' FOIM calculation. It can be interpreted as `mean(psi)`.

With the dev branch we could see that the equilibrium EIR wasn't held:

```r
parameters <- get_parameters(list(human_population=10000, individual_mosquitoes=FALSE))
parameters <- set_equilibrium(parameters, init_EIR=20)
output <- run_simulation(365 * 5, parameters)
ggplot(output) + geom_line(aes(timestep, EIR * 365))
ggplot(output) + geom_line(aes(timestep, FOIM_1))
```

eir_pre
foim_pre

After the FOIM is normalised we hold a more convincing equilibrium...

eir_post
foim_post

### DIFF
--- a/R/biting_process.R
+++ b/R/biting_process.R
@@ -110,7 +110,13 @@ simulate_bites <- function(
 
     renderer$render(
       paste0('lambda_', parameters$species[[s_i]]),
-      mean(lambda),
+      sum(lambda),
+      timestep
+    )
+
+    renderer$render(
+      paste0('normal_lambda_', parameters$species[[s_i]]),
+      sum(lambda) / mean(psi),
       timestep
     )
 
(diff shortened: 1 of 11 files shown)


htopazian (approved):
Nice Giovanni, thanks for investigating this

PR #122, opened 2021-07-05
Fix prophylaxis curve to be weibull survival

### LINKED ISSUE #121: The form of the drug prophylaxis curve
We were discussing parameterising the prophylaxis curves for the different drug types in the model. We noticed that in Okell et al. 2014 Nat Comms, the form of the prophylaxis curve is a Weibull survival function: P=exp((-t/lambda)^w), where lambda is the scale parameter and w is the shape parameter. In the malariasimulation model (human_infection.R), the prophylaxis curve is the Weibull pdf (dweibull) instead of the survival curve. This seems to imply a period of low (zero) prophylaxis at times immediately after treatment, with prophylaxis taking effect as the mean of the distribution is approached. We wanted to check this, and suggest that the survival curve should be used instead of the weibull pdf? Thus, prophylaxis is initially at its maximum level (directly after treatment) and then declines monotonically. 

Lucy has given us the following weibull survival curve parameterisations (for the two ACT drugs):
time<-0:50

scale_al<-10.6
shape_al<-11.3

scale_dp<-28.1
shape_dp<-4.4

p_protect_al<- exp(-((time/scale_al)^shape_al))
p_protect_dp<- exp(-((time/scale_dp)^shape_dp))

plot(time,p_protect_dp,type="l", ylab="probability of protection")
lines(time,p_protect_al,col="blue")

Thanks!

giovannic:
Closes #121


pahanc (approved):
Thanks Giovanni that looks good to me! In addition, we have another drug parameterisation that we would like to add to the model, but I will follow up with you separately on that.

PR #129, opened 2021-08-11
Fix hanging ODE for low rainfall

giovannic:
An ODE parameter (carrying capacity) was flipping between 0 and a few million within a fraction of a second. When we have a large carrying capacity and rainfall falls to 0, we get a stark increase in larval mortality.

We fix it with a minimum rainfall level of .01.

Changes:

 * Errors and debug output for too much work
 * Test that rainfall doesn't drop to  0
 * Update CI script


src/adult_mosquito_ode.h
```diff
@@ -24,7 +24,7 @@ struct AdultMosquitoModel {
     double mu; //death rate for adult female mosquitoes
     const double tau; //extrinsic incubation period
     double foim; //force of infection towards mosquitoes
-    AdultMosquitoModel(MosquitoModel, double, double, double);
+    AdultMosquitoModel(MosquitoModel, double, double, double, double);
```
richfitz, line 27: should this be `int` or `size_t` here for max steps?

src/mosquito_biology.cpp
```diff
@@ -20,7 +20,7 @@ double carrying_capacity(
 ) {
     if (model_seasonality) {
         double r = rainfall(timestep, g0, g, h);
-        return std::max(K0 * r, .01) / R_bar;
+        return std::max(K0 * r / R_bar, .01);
```
richfitz, line 23: worth pulling the `0.01` magic number into a const or `#define`?

src/mosquito_biology.cpp
```diff
@@ -45,8 +45,8 @@ double rainfall(
     double result = g0;
     for (auto i = 0u; i < g.size(); ++i) {
         result +=
-            g[i] * cos(2 * M_PI * t * (i + 1) / 365) +
-            h[i] * sin(2 * M_PI * t * (i + 1) / 365);
+            g[i] * cos(2 * M_PI * (t % 365) * (i + 1) / 365) +
+            h[i] * sin(2 * M_PI * (t % 365) * (i + 1) / 365);
     }
-    return std::max(result, 0.);
+    return std::max(result, .01);
```
richfitz, line 51: Is this the same 0.01 as above?

src/mosquito_biology.cpp
```diff
@@ -45,8 +45,8 @@ double rainfall(
     double result = g0;
     for (auto i = 0u; i < g.size(); ++i) {
         result +=
-            g[i] * cos(2 * M_PI * t * (i + 1) / 365) +
-            h[i] * sin(2 * M_PI * t * (i + 1) / 365);
+            g[i] * cos(2 * M_PI * (t % 365) * (i + 1) / 365) +
```
richfitz, line 48: are these modulos needed?

src/adult_mosquito_ode.h
richfitz: I see you use `size_t` below in `src/mosquito_ode.cpp`

richfitz (commented):
A few small comments here. I'm also not clear (given the explanation above) why foim is now passed to the constructor, but that's just unfamiliarity with the model

src/adult_mosquito_ode.h
giovannic: Ah, it's a bit confusing because there are several classes at play here. I've made an issue to improve this documentation #132 

The `size_t` below is for the `Solver`. Whereas the `double` here is for the `AdultMosquitoModel`.

We've decoupled the model from the solver, to help switch between individual mosquitoes and compartmental ones.

src/mosquito_biology.cpp
giovannic: Thanks, it was an unnecessary leftover from debugging

giovannic:
> A few small comments here. I'm also not clear (given the explanation above) why foim is now passed to the constructor, but that's just unfamiliarity with the model

So we have these `*MosquitoModel` classes (configured based on if adult mosquitoes are individual or not) which store the parameters for the differential equations. We update them with foim for the current timestep. Coincidentally, this is done _before_ the solver tries to access foim for that timestep.

In trying to recreate the issue, I had to construct a MosquitoModel, then update it with an foim, before jumping the solver to the time of interest. It's much neater to construct a MosquitoModel that's ready to go.

richfitz (approved):

PR #147, opened 2021-10-11
Bug/severe

giovannic:
This PR adds p_inc, p_detect outputs. These are more consistent with Jamie's outputs and smoother.

The p_inc/p_detect outputs sum the probabilities of infection directly into the output. The existing n_inc/n_detect outputs the number of individuals which were sampled for infection. The existing approach comes with more volatility for smaller a population size, which may not be necessary for a simple analyses.

These are some example outputs...

```r
p <- malariasimulation::get_parameters(list(
  severe_enabled=TRUE,
  severe_incidence_rendering_min_ages=0,
  severe_incidence_rendering_max_ages=36500,
  human_population = 10000,
  v=0,
  fvt=0
))
p <- malariasimulation::set_equilibrium(p, 100)
o <- malariasimulation::run_simulation(365 * 10, p)

# plot p_detect vs n_detect
ggplot2::ggplot(o) + ggplot2::geom_line(aes(x = timestep, y = p_detect_730_3650))
ggplot2::ggplot(o) + ggplot2::geom_line(aes(x = timestep, y = n_detect_730_3650))

# plot p_inc_severe vs n_inc_severe (aggregated by year)
df <- data.frame(x = seq(10), y = colSums(matrix(o$p_inc_severe_0_36500, ncol = 10)))
ggplot(df) + geom_line(aes(x = x, y = y)) + xlab('year') + ylab('p_inc_sev')
df <- data.frame(x = seq(10), y = colSums(matrix(o$n_inc_severe_0_36500, ncol = 10)))
ggplot(df) + geom_line(aes(x = x, y = y)) + xlab('year') + ylab('n_inc_sev')
```
pdet
ndet
psev
nsev

Summary of changes:
 * severe mortality fixes
 * refactored incidence rendering (one function for clinical and severe)
 * p_ outputs for detect and inc
 * update test cases
 * bitset_index for speed
 * documentation updates
 * remove obsolete C++ test helpers


R/model.R
```diff
@@ -31,18 +31,31 @@
 #'  * n_detect: number of humans with an infection detectable by microscopy between an inclusive age range at this timestep. This
 #' defaults to n_detect_730_3650. Other age ranges can be set with
 #' prevalence_rendering_min_ages and prevalence_rendering_max_ages parameters.
-#'  * n_severe: number of humans with a severe infection detectable by microscopy 
-#'  between an inclusive age range at this timestep. Age ranges can be set with
+#'  * p_detect: the sum of probabilities of detection by microscopy between an
```
pwinskill, line 34: As the output now contains both, I wonder if it will help users to have an example (in vignette?) when to use one or the other. The nuance of these output types was not something I really remember clocking before, so some guidance will likely be needed.
I guess that users would in general use the smoother p_inc/p_detect outputs, unless they were trying to model "actual" population numbers or elimination questions - is that how you see it too?

pwinskill (commented):
Interesting!
Not a part of the model I'd looked into in a lot of detail for quite a while. A couple qus/comments here

R/model.R
pwinskill: I think potentially useful yes and worth keeping seeing as they have been coded up already.

That output is very helpful. I think that with a litte explainer of how/why the black and grey differ and when to use which kind of output would be great.

pwinskill (approved):

PR #154, opened 2022-01-07
Feat/demography

giovannic:
This PR implements a custom demography for a constant population.

It would be cool to see if the Demography.Rmd vignette makes sense, or if it could be explained more clearly.


vignettes/Demography.Rmd
```diff
@@ -0,0 +1,139 @@
...
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+```{r setup}
+suppressPackageStartupMessages(library(ggplot2))
```
htopazian, line 18: ggplot??!?!?!?! yes!

vignettes/Demography.Rmd
```diff
@@ -0,0 +1,139 @@
...
+
+# Parameterisation
+
+We are going to create a scenario where we track severe prevalence for 5 years.
+
+```{r}
+year <- 365
+month <- 30
+sim_length <- 5 * year
+human_population <- 1000
```
htopazian, line 30: this is a great example - super clear.

R/parameters.R
```diff
@@ -405,7 +407,9 @@ get_parameters <- function(overrides = list()) {
     severe_incidence_rendering_min_ages = numeric(0),
     severe_incidence_rendering_max_ages = numeric(0),
     # misc
+    custom_demography = FALSE,
```
htopazian, line 410: I think we should reference the average_age variable here (line #91). If `custom_demography = FALSE` then the model uses average_age = 7663 which is the exponential distribution, right? So does `custom_demography = TRUE` override the average_age argument?

vignettes/Demography.Rmd
giovannic: 😄 we had ggplot all along in DESCRIPTION! Not the whole of tidyverse though. I could remove it? 😛

PR #158, opened 2022-02-09
Bug/immunity incidence

pwinskill:
A small adjustment to the estimation of immunity blood-stage, clinical and severe disease to align incidence and severe incidence estimates.

Immunity values >0 and now evaluated +0.5. This is likely to align with the mid-point of the range that would have been used for the initial fitting.

Note, this does not apply to maternally-aquired immunity.

Adjustments also made to immunity function testing. Although changes were not picked up by some of those tests due to the specified tolerances not being accurate enough. @giovannic  I upped the tolereance by an order of magnitude but we might want to push this further given how sensitive the modelled incidence is to these values?

### DIFF
--- a/R/human_infection.R
+++ b/R/human_infection.R
@@ -376,6 +376,8 @@ boost_immunity <- function(
 
 # Implemented from Winskill 2017 - Supplementary Information page 4
 clinical_immunity <- function(acquired_immunity, maternal_immunity, parameters) {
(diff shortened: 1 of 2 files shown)


pwinskill:
Not sure why the vignette build is failing on macOS - have you seen that one before?

giovannic:
Thanks Pete this looks great!

On the MacOS test, I've made an issue for it #159. It's not new :/

giovannic (approved):

PR #172, opened 2022-07-21
Add ITN usage renderer

giovannic:
ITN usage statistics


R/vector_control.R
```diff
@@ -185,3 +185,13 @@ bednet_decay <- function(t, gamma) {
 spraying_decay <- function(t, theta, gamma) {
   1 / (1 + exp(-(theta + gamma * t)))
 }
+
+net_usage_renderer <- function(net_time, renderer) {
+  function(t) {
+    renderer$render(
+      'net_usage',
```
pwinskill, line 192: Net usage in general would be a proportion.
To avoid confusion and keep with the naming convention for other outputs (e.g. prevalence), do you think calling this output `n_use_net` would be a good idea?

pwinskill (commented):
Thanks @giovannic !
One small comment, but works great for me 😎

pwinskill (approved):
❤

PR #177, opened 2022-07-21
Feat/time vary mda age

### LINKED ISSUE #173: Time-varying min and max age groups from MDA/SMC
Make sure old arguments are renamed and/or inputs validated with an informative error as this would be a breaking change

pwinskill:
Implements time-varying min and max ages for MDA, and by extension SMC.

`min_age` and `max_age` arguments have been renamed to `min_ages` and `max_ages`, although R will match the arguments with partial name matching. So, to safeguard there are now checks throwing an error if these arguments are not vectors of the same length as the coverage vector inpur.

**Breaking change** - scalar values for the `min_age` and `max_age` arguments will no longer work, unless there is a single coverage timepoint, which is then I think the desired behaviour.

closes #173


giovannic (approved):
Looks great!

PR #188, opened 2022-08-09
ITN and IRS survive rebirth

giovannic:


pwinskill:
Just to confirm, we do need both the ITN surviving re-birth and the schedule clearance before scheduling to get expected behaviour. Here is the comparison with just rebirth (black analytical, green malariasimulation):
image
and here with both rebirth and altered scheduling:
image
I think you need to fully overwrite when resampling from the exponential otherwise I think with every new net dist you have the possibility of sampling and queuing up a more recent net throw time.

pwinskill (approved):
Thanks @giovannic - looks good!
I wonder if it is worth checking through that this re-scheduling issue doesn't pop up anywhere else too?

giovannic:
> Thanks @giovannic - looks good! I wonder if it is worth checking through that this re-scheduling issue doesn't pop up anywhere else too?

throw_away_net is the last targeted event now. We threw them all away for processes

PR #190, opened 2022-08-16
Rendering for number of persons treated for clinical malaria or via MDA, SMC and PMC

### LINKED ISSUE #189: Rendering of n_treated is actually n_successfully_treated
This distinction might be important if using simulation outputs to cost a programme.
The cost would be the number treated (whether successful or not) multiplied by the drug cost, no the number successfully treated.

pwinskill:
I believe that the `n_treated` outputs are mostly used to cost intervention programmes, it is therefore the number of treatments given, not the number of effective treatments given that should be output. This PR moves the rendering of `n_treated` outputs for clinical treatment, MDA, SMC and PMC to be prior to the drug-efficacy draw.

The PR also adds a distinction between MDA and SMC outputs which are now `n_mda_treated` and `n_smc_treated` respectively.

Fixes #189 

Please do let me know if you have heard of a use-case where users need the output after the drug effiacy draw.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: malariasimulation
 Title: An individual based model for malaria
-Version: 1.4.1
+Version: 1.4.2
 Authors@R: c(
   person(
     given = "Giovanni",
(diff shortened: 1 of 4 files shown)


giovannic (approved):
Great catch on the mda rendering. Thanks!

PR #194, opened 2022-10-03
Reduce mixed transmission with testing

giovannic:
These changes add a `mixing_rdt` model parameter. When set to TRUE, the metapopulation model will reduce the transmission from external populations by the complement of the RDT prevalence in the external population.

Other parameters `rdt_intercept` and `rdt_coeff` should probably remain unchanged, they are set from Wu et al. 2015

Changes:

 * Add rdt testing parameters
 * Update biting_process to consider rdt testing for mixing
 * Add integration tests for rdt prevalence calculation and mixing reductions

### DIFF
--- a/R/biting_process.R
+++ b/R/biting_process.R
@@ -112,6 +112,10 @@ simulate_bites <- function(
     a <- .human_blood_meal_rate(f, s_i, W, parameters)
     lambda <- effective_biting_rates(a, .pi, p_bitten)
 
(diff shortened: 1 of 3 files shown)


htopazian (commented):
I've installed a local copy and run some tests with printing the p_captured_t, rdt_positive, p_success, and test_and_treat_coeff.
Setting rdt_itercept = 0 and rdt_coeff = 1 doesn't seem to turn off the rdt_positive part. 
Other things look okay, I can't find the problem!

PR #219, opened 2023-02-15
Feat/parameter draws

pwinskill:
Adds in the option for the user to overwrite default model parameters (median estimates from joint posterior of MCMC fitting), with one of 1000 draws from the joint posterior.

Example use:
```
# Get the default parameters
p <- malariasimulation::get_parameters()
# Overwrite the parameters with draw number 1:
p_draw <- p |>
  set_parameter_draw(1)
```

Original 1000 parameter draw files from fitting were processed to:
1. Only keep parmeters that varied between draws
2. Rename to be aligned with malariasimulation inputs

Following conversion I checked against the malaria simulation defaults to ensure the conversion was correct. Draw distributions v malariasimulation default value:
Comparison_plot
Draw medians v malariasimulation default value:
Comparison_median_plot

In the help for `set_parmeter_draw()` I've flagged that it needs to be set prior to calling `set_equilibrium()`

Here's an example of plotting PfPr for 30 random draws (and the default in black)
```
# Default pars
p <- malariasimulation::get_parameters(overrides = list(human_population = 4000)) |>
  set_equilibrium(init_EIR = 5)
r1 <- run_simulation(365 * 5, p)

plot(r1$timestep, r1$n_detect_730_3650 / r1$n_730_3650, t = "l", ylim = c(0, 1), ylab = "PfPr", xlab = "T")

# 30 Random draws
cols <- rainbow(30)
for(i in 1:30){
  # Overwrite the parameters with draw number 1:
  p_draw <- malariasimulation::get_parameters(overrides = list(human_population = 4000)) |>
    set_parameter_draw(sample(1:1000, 1)) |>
    set_equilibrium(init_EIR = 5)
  r2 <- run_simulation(365 * 5, p_draw)
  lines(r2$timestep, r2$n_detect_730_3650 / r2$n_730_3650, col = cols[i])
}
lines(r1$timestep, r1$n_detect_730_3650 / r1$n_730_3650, col = "black")
```
pfpr_draw


R/parameters.R
```diff
@@ -517,3 +517,23 @@ parameterise_total_M <- function(parameters, total_M) {
...
   parameters
 }
+
+#' Use parameter draw from the join posterior
+#' 
+#' Overrides default (median) model parameters with a single draw from the fitted
+#' joint posterior. Must be called prior to set_equilibrium.
+#'
+#' @param parameters the model parameters
+#' @param draw the draw to use. Must be in integer between 1 and 1000
```
htopazian, line 527: "... must be an integer between ..."

htopazian (changes requested):
The R/sysdata.rda file name is a little misleading - can it be renamed to something indicating that this is the parameter draws set? 

In R/parameters.R, the parameter da (dur_A) should be changed to 195 instead of 200 (both default value and in documentation). This is the correct value from the Presidents Initiative supplement. 

Can we add a small vignette with an example? Should include explanation for new users on what we mean by parameter draws (where do these come from). What general model variables or concepts are affected by changing the draw number. And the pros / cons of selecting a random sample of draws each time - should this be randomly selected one time at the start of an analysis and then held constant for each time a user re-runs their code, or should this be random each time? 

Verified that 32 parameters selected are the ones which change between runs :white_check_mark:

Verified that function works as anticipated :white_check_mark:

I really like this function, it will make life much easier and eliminate user errors in copying over parameter draws from the old model.

R/parameters.R
pwinskill: Thanks @htopazian 🙏,

Have fixed the typo.
Regarding the R/sysdata.rda file I just followed the steps [here](https://r-pkgs.org/data.html#sec-data-sysdata) for internal data in a package, so I'm not sure I can rename it?

Agree about the vignette! I was going to ask the vignette update squad to have a go at it  - I can add details to an issue to make sure it doesn't get forgotten

giovannic (changes requested):
I agree with Hillary, sysdata.rda looks mysterious.

It would feel more maintainable if the data file was stored in a data directory. And the parameter_draw variable in the global environment was documented somewhere.

This seems like a good way to do it: https://kbroman.org/pkg_primer/pages/data.html

R/parameters.R
pwinskill: Regarding dur_A, in the original params files I pulled out it is 195 - so I expect that the PMI paper SI documentation is just a mistake/rounding

giovannic: Just seen this, I didn't know the `data` directory had public visibility which would be a drawback...

giovannic: Just updated my comment for this

htopazian: Wait for dur_A, the default in malariasim is 200 https://github.com/mrc-ide/malariasimulation/blob/master/R/parameters.R#L240, but I think this should be 195 to match the original params files and the PMI documentation (I think the SI is correct).

giovannic: Ah, I set it to line up with the [legacy model](https://github.com/mrc-ide/Malaria_model/blob/master/input_files/model_parms.txt#L71)

giovannic:
> the script just reads in and renames a bunch of txt files from the original fitting.

I think that would be really useful. So that we could see the names and data in a human readable format. And if we want to update the draws, we could simply update the txt file, run the script, and add to the repo. The changes would even be recognised by git!

pwinskill:
> > the script just reads in and renames a bunch of txt files from the original fitting.
> 
> I think that would be really useful. So that we could see the names and data in a human readable format. And if we want to update the draws, we could simply update the txt file, run the script, and add to the repo. The changes would even be recognised by git!

Thanks for the pointers @giovannic - have added this in now as well

giovannic (approved):
Thanks!

htopazian (approved):
Nice!

Issue #221, opened 2023-02-17
Create a vignette for parameter draws


pwinskill:
This vignette would guide the user through the additional functionality detailed in [this PR](https://github.com/mrc-ide/malariasimulation/pull/219).

Some more discussion of the vignette [here](https://github.com/mrc-ide/malariasimulation/pull/219#pullrequestreview-1301128939)

lmhaile:
Hi Pete, I can work on this if useful! I have some code related to the linked PR I can adapt.

pwinskill:
Perfect - thank you @lmhaile!

PR #225, opened 2023-03-02
Vignettes/variation

kellymccain28:
Updates to the variation vignette include: 

- Removal of ggplot2 dependency 
- Removal of references to p_detect (I was not sure what the purpose of including the distinction between n_detect and p_detect was, so I removed it to focus on the random variation in the model based on population size, but happy to add it back in.)
- Updating and adding explanations/slight reorganizing


vignettes/Variation.Rmd
```diff
@@ -1,66 +1,90 @@
...
+# A larger population
+output_big_pop <- run_simulation(365, simparams)
+
+# Plotting 
+par(mfrow = c(1,2))
+plot_n_detect(output_small_pop); title('n_detect at n = 100')
+plot_n_detect(output_big_pop, ylab = FALSE); title('n_detect at n = 10,000')
+```
 
 ## Estimating variation
```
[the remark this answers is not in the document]
pwinskill: To add to this again 😂...
I'm wondering perhaps if this should be a section in a larger vignette that also details the parameter draw uncertainty adressing: https://github.com/mrc-ide/malariasimulation/issues/221

pwinskill (changes requested):
Nice start @kellymccain28 👌.
A few specific changes requested and also a comment about if we want to include this as part of a larger vignette including the parameter draw uncertatinty

vignettes/Variation.Rmd
```diff
@@ -1,66 +1,90 @@
...
+`malariasimulation` is a stochastic, individual-based model and, as such, simulations run with identical parameterisations will generate non-identical, sometimes highly variable, outputs. To illustrate this, we will compare the prevalence of malaria over a year in simulations with a small and a larger population, where a smaller population is more likely to have wide variations between simulation runs. 
 
-Malariasimulation is a stochastic model, so there will be be variation in model outputs. For example, we could look at detected malaria cases over a year...
+### Plotting functions
+First, we will create a few plotting functions to visualise outputs. 
+```{r}
+cols <- c("#E69F00", "#56B4E9", "#009E73", 
+          "#F0E442", "#0072B2", "#D55E00", "#CC79A7")
+          
+plot_n_detect <- function(output, ylab=TRUE){
```
htopazian, line 22: extrememly nitpicky, but could we keep the same spacing rules and stylistic things throughout all of the vignettes? E.g. `ylab = TRUE` vs. `ylab=TRUE`, space after comma vs. not, double " vs single ', starting with a capital vs lowercase letter after the # sign, etc.

vignettes/Variation.Rmd
```diff
@@ -1,66 +1,90 @@
...
+  plot(x = output$timestep, y = output$n_detect_730_3650, 
+       type = 'l', col = cols[3],
+       xlab = 'Time (days)', ylab = ylab,
+       xaxs = "i", yaxs = "i", bty = "l")
+  grid(lty = 2, col = 'grey80', lwd = 1)
+}
+```
 
+## Parameterisation
+First, we will use the `get_parameters()` function to generate a list of parameters, accepting the default values, for two different population sizes. Then we will update the parameters to match equilibrium parameters and to achieve the initial EIR using `set_equilibrium()`
```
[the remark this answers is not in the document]
htopazian: ... "the only parameter which changes between the two parameter sets is the argument for `human_population`"

vignettes/Variation.Rmd
```diff
@@ -1,66 +1,90 @@
...
+
+simparams <- set_equilibrium(simparams, init_EIR = 20)
 ```
 
-The n_detect output shows the result of sampling individuals who would be detected by microscopy.
 
-While the p_detect output shows the sum of probabilities of detection in the population.
+## Simulations
+
+The `n_detect_730_3650` output below shows the total number of individuals in the age group rendered (here, 730-3650 timesteps or 2-10 years) who have microscopy-detectable malaria. Notice that the output is smoother at a higher population.
```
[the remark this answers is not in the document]
htopazian: Yes agree, showing prevalence might be a more useful metric. Dividing n_detect by n

vignettes/Variation.Rmd
```diff
@@ -1,66 +1,90 @@
...
+# A small population
+output_small_pop <- run_simulation(365, simparams)
+
+# A larger population
+output_big_pop <- run_simulation(365, simparams)
+
+# Plotting 
+par(mfrow = c(1,2))
+plot_n_detect(output_small_pop); title('n_detect at n = 100')
+plot_n_detect(output_big_pop, ylab = FALSE); title('n_detect at n = 10,000')
```
htopazian, line 69: Here it is a little hard to see the difference between the two plots immediately - I think changing to a prevalence output will help, and you could also force the y axes to cover the same range between both plots so it is more obvious and we are comparing same to same

vignettes/Variation.Rmd
```diff
@@ -1,66 +1,90 @@
...
 
-```{r}
+```{r, fig.align='center',fig.height=4, fig.width=10}
 outputs <- run_simulation_with_repetitions(
-  timesteps = 365,
-  repetitions = 10,
-  overrides = p,
-  parallel=TRUE
+  timesteps = 365, # We will run each simulation for 365 timesteps.
+  repetitions = 10, # The simulation will be run 10 times
```
htopazian, line 79: Use "simulation" for either the overall run or each individual run. 
`# Each repetition will be run for 365 days`
`# The simulation will be run 10 times`

vignettes/Variation.Rmd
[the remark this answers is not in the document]
htopazian: I think this is useful as it is a function available in `malariasimulation`. And if someone is using the package outside of Imperial to do some small test cases they may want to just use repetitions instead of parameter draws. But I agree with Pete that we should add more text to clarify the differences between repetitions vs. increasing population size vs. parameter draws. I think it is okay to keep here for now so as not to clutter the parameter draws vignette. But to be clear that while running with repetitions is an option that we are illustrating, it is preferable to increase population size or use parameter draws.

htopazian (changes requested):
Hi Kelly, nice work (esp in figuring out all that base plotting code ...). Comments and a few responses to Pete :)

vignettes/Variation.Rmd
```diff
@@ -1,91 +1,151 @@
...
-We can estimate the variation in the number of detectable cases by repeating the simulation several times...
+```{r, fig.align="center",fig.height=6, fig.width=8}
+# A small population
+output_small_pop <- run_simulation(timesteps = 365, parameters = simparams_small)
 
-```{r}
+# A larger population
+output_big_pop <- run_simulation(timesteps = 365, parameters = simparams_big)
+
+# Plotting
```
pwinskill, line 83: The stochastic variation would be much more obvious if you zoom the y axis limits to fit more closely to the data

vignettes/Variation.Rmd
```diff
@@ -1,91 +1,151 @@
...
 
-```{r setup, include=FALSE}
+```{r setup}
 library(malariasimulation)
-library(cowplot)
-library(ggplot2)
 ```
 
-## Variation in outputs
+`malariasimulation` is a stochastic, individual-based model and, as such, simulations run with identical parameterisations will generate non-identical, sometimes highly variable, outputs. To illustrate this, we will compare the prevalence and incidence of malaria over a year in simulations with a small and a larger population, where a smaller population and the incidence measure are both more likely to have wide variations between simulation runs.
```
pwinskill, line 14: The last sentence doesn't make sense to me?

vignettes/Variation.Rmd
```diff
@@ -1,91 +1,151 @@
...
+
+# Plotting 
+par(mfrow = c(2,2))
+plot_prev(output_small_pop); title("Prevalence at n = 1,000")
+plot_inci(output_small_pop); title("Incidence at n = 1,000")
+plot_prev(output_big_pop, ylab = FALSE); title("Prevalence at n = 10,000")
+plot_inci(output_big_pop); title("Incidence at n = 10,000")
+```
+
+## Estimating stochastic variation
```
pwinskill, line 91: I'm still not sold on the value of this final section.

More useful, in my opinion, would be to show an example of stochatic fade out (elimination) in a small population, compared with stable transmission in the larger population. That would need 2 runs with the same low eir and different pop sizes - you'd need to set the seed to ensure the user saw the same output each time

vignettes/Variation.Rmd
```diff
@@ -1,91 +1,151 @@
...
+        col = "grey90", border= "grey95")
+lines(x = output$timestep, y = output$median, col = cols[3], lwd = 2)
+grid(lty = 2, col = "grey80", lwd = 1)
+title("IQR spread")
+legend("bottomleft", legend = c("Median","IQR Spread"), col = c(cols[3], "grey90"),
+       lwd = 5, xpd = TRUE, horiz = TRUE, cex = 1, seg.len = 1, bty = "n")
+
+xx <- c(output$timestep, rev(output$timestep))
+yy <- c(output$lowci, rev(output$highci))
+plot(x = output$timestep, y = output$mean,
```
pwinskill, line 141: These plots are empty for me

pwinskill (changes requested):
A few more comments from me :)

pwinskill (changes requested):
Hi @kellymccain28,

Looking good. In the final plot it looks like PfPR starts at 0
image

This can't be correct otherwise there wouldn't be any transmission later on. Can you double check if it is a processing/plotting issue? It might help to plot those last two PfPr plots on a ylim scale of (0, 0.2) to better see what is going on

kellymccain28:
For the final plot, I changed the seed so it starts at >0% prevalence and reduced the y axis limits to c(0,0.2) as suggested. THere is a point in the plot here where the number of cases detected is at 0 for a timestep then goes back up to >0 which I think makes sense as there will still be infecitous mosquitoes buzzing around that could infect people
image

pwinskill (approved):

PR #235, opened 2023-04-25
Feat/carrying capacity

pwinskill:
Adds in functionality to allow completely custom time varying and species specific carrying capacity vectors.

Functionality is implemented to

Provide the user with the calculation of the baseline carrying capacity for each species `get_init_carrying_capacity()`
Allow species specific carrying capacity change to be made at any number of user-given timepoints using `set_carrying_capacity()`

Examples of use and implementation are included in the vignette `"Modifying the carrying capacity"`.

Other changes implemented:

- Modifications and renaming of the History object -> Timeseries.
- More information in the solve error message if too much work


src/aquatic_mosquito_eqs.cpp
```diff
@@ -9,73 +9,81 @@
...
+  return [&model](const state_t& x, state_t& dxdt, double t) {
+    auto K = carrying_capacity(
+      t,
+      model.model_seasonality,
+      model.g0,
+      model.g,
+      model.h,
+      model.K0[std::round(t)],
+              model.R_bar,
+              model.rainfall_floor
```
giovannic, line 21: indentation

giovannic (changes requested):
Ok!

Maintenance wise, I think it would be easier to have one `set_carrying_capacity` function. And we could demonstrate how to use it in the different ways in the vignette.

pwinskill:
Thanks for the comments @giovannic 🙏. A few thoughts

**Regarding a single `set_carrying_cacpacity()` function** 
Agree this would be easier to maintain. This would be different to what we are currently doing with seasonality. As I see it currerntly the steps are:

1. User (effectively) sets the carrying capacity using `set_equilibrium()`
2. User specifies changes to carrying capacity using `model_seasonality = TRUE` etc
3. Modifications are then implemented within the model code.

It is that flow that I was trying to maintain with the proposed changes, addinging additional functions to specify unqiue and independent modifications to the carrying capacity.

**Regarding iterpolation**

So, unsurprisingly, it is possible to make the solver fall over with these changes. This occurs with large step-changes to the carrying capacity. I think the "flexibility" of a completly free `set_carrying_capacity()` function is the challenge here, as it will be impossible to tell _a priori_ if large step-like changes from the user are appropriate (for example implementing larval source management) or just clunky input.
When searching for parameterisations that would cause the solver to fail it was only when I put in what I would interpret as unreasonable parmeterisations (from a biological perspective) that issues would occur. In this case it seems completly reasonable to me that the model would(should) fall over. If that is the case then it would seem reasonable to me to inform the user about this via the vignette (and perhaps give a few more clues with the solver error message).

giovannic (approved):
Thank you so much for this!

PR #244, opened 2023-06-06
Changed mum parameter for consistency with single vector species (gam…

RJSheppard:
…biae, rather than current: 'All')


htopazian (approved):
Nice catch Richard, looking good! We'll need to remember to catch these for the vignettes - can rebase the vignettes branch to this `dev` version once Tom's is merged in

pwinskill (approved):
🙏

PR #245, opened 2023-06-16
Vivax parameters

RJSheppard:
I made some changes to the parameters.R file (the documentation didn't quite match up with the order and sectioning). I've made it a bit more consistent. I can create a pull request into dev if you think that's worth it (thoughts/feedback welcome!).

I also created a function to change the parameters to match what is needed to vivax. Currently, the lines removing unnecessary parameters are commented out because those changes haven't been made yet (and taking them out now would break everything).

Also had a quick go at writing some basic tests.

Any suggestions in formatting the documentation are welcome!


R/parasite_parameters.R
```diff
@@ -0,0 +1,233 @@
...
+#' @export
+set_parasite <- function(parameters, parasite) {
+  if (length(parasite) != 1) {
+    stop('Only one parasite species can be selected')
+  }
+  if (!parasite %in% c("vivax","falciparum")){
+    stop('Parasite species must be "vivax" or "falciparum"')
+  }
+  if (parasite == "falciparum"){
+    return(parameters)
```
giovannic, line 111: does this method work if I do this?

```r
parameters <- set_parasite(parameters, 'vivax')
parameters <- set_parasite(parameters, 'falciparum')
```

I think we should either restrict this function to only go from falciparum -> vivax by renaming it `set_vivax`, or support going from vivax parameters to falciparum ones (probably less likely).

R/parasite_parameters.R
```diff
@@ -0,0 +1,233 @@
...
+    
+    # Under misc
+    parameters$parasite <- "vivax"
+    
+    ## fixed state transitions
+    # dd is the same
+    parameters$dt <- 1
+    parameters$da <- 10
+    # du is now an equation
+    # parameters <- parameters[names(parameters) != "du"]
```
giovannic, line 124: Why is this commented out?

Perhaps we could have a separate piece of logic at the bottom to remove all unused parameters:

```r
unused parameters <- c(
  # immunity parameters
  'du',
  # decay rates
  '...'
)
parameters <- parameters[!names(parameters) %in% unused_parameters]
```

R/parasite_parameters.R
```diff
@@ -0,0 +1,233 @@
+#' @title Parameterise the parasite species to use in the model
+#'
```
giovannic, line 2: Could we have a citation somewhere in this description. It's useful when others want to write up their S.Is

R/parasite_parameters.R
```diff
@@ -0,0 +1,233 @@
...
+    # Under misc
+    parameters$parasite <- "vivax"
+    
+    ## fixed state transitions
+    # dd is the same
+    parameters$dt <- 1
+    parameters$da <- 10
+    # du is now an equation
+    # parameters <- parameters[names(parameters) != "du"]
+    # del, dl, dpl, mup and mum are the same
```
giovannic, line 125: [nit] it feels like a lot of information to comment on which parameters stay the same

tests/testthat/test-vivax.R
```diff
@@ -0,0 +1,12 @@
+test_that('Test model runs using falciparum', {
+  parameters <- get_parameters()
+  sim_res <- run_simulation(timesteps = 100, parameters = parameters)
+  expect_equal(nrow(sim_res), 100)
+})
```
giovannic, line 5: This feels duplicated from test-simulation-e2e.R
```suggestion
```

giovannic (changes requested):
This looks like a great place to start!

The issue for me is that we don't want to commit commented out code. Because then we have to remember to uncomment it. And it looks like it's in several different places.

I'm guessing they are commented out because it breaks the end-to-end (e2e) test to see that a whole vivax simulation could run, but we haven't implemented that yet.

I would consider a simpler tests at this stage to check that `set_parasite` returns a valid parameter list, that it doesn't have unnecessary keys, that parasite is set to "vivax"...

R/parasite_parameters.R
RJSheppard: I wanted to keep track of all the parameters to make sure I wasn't ignoring any, but can maybe put that information into a vignette or something.

R/parasite_parameters.R
RJSheppard: Good point! I've added a citation.

R/parameters.R
```diff
@@ -207,96 +249,60 @@
...
+  # severe disease immunity parameters
+  # asymptomatic immunity parameters
+  # immunity boost grace periods
+  # infectivity towards mosquitos
+  # unique biting rate
+  # delay for infection
+  # initial immunities
+  
+  parameters <- as.list(
+    parasite_parameters[parasite_parameters$parasite == parasite, !is.na(parasite_parameters[parasite_parameters$parasite == parasite,])])
```
pwinskill, line 290: You could pre-do this when preparing the data in data-raw/parasite_parameters.R
In the above script why not create something like:
```
parasites <- c("falciparum", "vivax")
parasite_parameters <-  lapply(parasites, function(parasite){
  as.list(
    parasite_parameters[parasite_parameters$parasite == parasite, !is.na(parasite_parameters[parasite_parameters$parasite == parasite,])]
    )
})
names(parasite_parameters) <- parasites
```
and save that as the data object. Then when called in `get_parameters()` you could just retrieve it with `parasite_parameters[[parasite]]`?

R/parameters.R
```diff
@@ -1,40 +1,55 @@
...
+#' interventions on Plasmodium vivax." by White, Michael T., et al.
 #'
 #' @param overrides a named list of parameter values to use instead of defaults
+#' @param parasite Plasmodium parasite species ("falciparum" or "vivax"); default = "falciparum"
+#' 
 #' The parameters are defined below.
 #'
 #' fixed state transitions:
 #'
 #' * dd - the delay for humans to move from state D to A; default = 5
```
pwinskill, line 20: If expanding here to multiple parasites, do we want parasite-related defaults to be listed in this documentation of parameters, or should the csv be the reference to see defaults?

data-raw/parasite_parameters.csv
```diff
@@ -0,0 +1,4 @@
+parasite,dd,dt,da,du,del,dl,dpl,mup,mum,du_max,du_min,ku,au50,rm,rvm,rb,rva,rc,rid,b0,b1,ib0,kb,b,phi0,phi1,ic0,kc,theta0,theta1,kv,fv0,av,gammav,iv0,fd0,ad,gammad,d0,d1,id0,kd,ub,uc,uv,ud,cd,gamma1,cu,ct,ca,a0,rho,sigma_squared,n_heterogeneity_groups,de,delay_gam,dem,pcm,pvm,init_ica,init_iva,init_icm,init_ivm,init_id,init_idm,init_ib,f,gammal
```
pwinskill, line 1: Not sure what vivax_tmp is?

pwinskill, line 1: If we go this route - would this csv be a good place to store the referecnes for the default parameters. You could image it set up in long form such as:

| parasite | parameter | default | reference |

tests/testthat/test-vivax.R
```diff
@@ -0,0 +1,26 @@
+test_that('Test falciparum switch produces same', {
+  parameters <- get_parameters()
+  parameters_fal <- get_parameters(parasite = "falciparum")
+  expect_equal(parameters, parameters_fal)
```
pwinskill, line 4: `expect_identical()` ?

tests/testthat/test-vivax.R
```diff
@@ -0,0 +1,26 @@
...
+})
+
+test_that('Test vivax model runs', {
+  vivax_parameters <- get_parameters(parasite = "vivax_tmp")
```
pwinskill, line 8: See previous comment: not sure what "vivax_tmp" is?

R/parameters.R
```diff
@@ -151,35 +180,40 @@
...
-#' * clinical_treatment_drugs - a vector of drugs that are available for
-#' clinically diseased (these values refer to the index in drug_* parameters); default = NULL, NULL, NULL
-#' * clinical_treatment_coverage - a vector of coverage values for each drug; default = NULL, NULL, NULL
+#' * clinical_treatment_drugs - a list of drugs that are available for
+#' clinically diseased (these values refer to the index in drug_* parameters); default = NULL
+#' * clinical_treatment_timesteps - a list of vectors giving timesteps at which the
+#' clinical treatment coverage changes (these values refer to the index in drug_* parameters); default = NULL
+#' * clinical_treatment_coverage - a list of vectors giving coverage values for each drug; default = NULL
 #'
 #' PEV parameters:
```
pwinskill, line 198: Just a note while I remeber - some interventions won't be available for vivax, so we will need a catch in the relevant set intervention functions to stop them being implemented if `parasite == "vivax"`

pwinskill (commented):
Thanks Richard 🙏
Some comments and suggestion from me.
We also need to think how this interacts with the `set_parameter_draw()` functionality as these are currently only available with Pf

data-raw/parasite_parameters.csv
RJSheppard: I had this thought too. I think we would need to be a little careful. I originally had the parameter data in long form, but because the values contain numerics and strings, they all revert to strings (which they don't do in wide form). Adding a reference to each column would have the same problem. There could be a way around this though...? The parameters also contain vectors and empty lists and numerics which I also couldn't work out how to include in a csv, so I gave up on trying to include them and settled with doing just the parasite parameters.

tests/testthat/test-vivax.R
RJSheppard: See above :) It currently also contains all falciparum parameters required to make the model work without the vivax functionality.

tests/testthat/test-vivax.R
RJSheppard: Good shout!

R/parameters.R
RJSheppard: Makes sense. I had a thought that at least the parameters for drug efficacy would need to be different. I guess we can talk about that later.

R/parasite_parameters.R
RJSheppard: I've completely reshifted the code, using a csv file for the parameters, and only loading the appropriate set in get_parameters. I think it saves a lot of space and faff.

R/parasite_parameters.R
RJSheppard: No longer needed.

R/parasite_parameters.R
RJSheppard: This is no longer an issue, but I've left in which parameters are used in the specific falciparum/vivax models

R/parameters.R
RJSheppard: I think we can change this later... if/when expanding can rethink as to how to document.

R/parameters.R
RJSheppard: Just added this in the commit I'm about to push.

RJSheppard:
I've reshaped the code with a parasite parameter csv file in a way that seems a bit tidier and avoids some of the problems highlighted. Let me know what you think!

giovannic (approved):

PR #246, opened 2023-06-16
Bug: order of initial state proportions inconsistent

RJSheppard:
I noticed that when setting the proportion parameters that the A and D initial proportions are swapped round in the results. There is a mismatch between the order in states and calculate_initial_counts. I think this has fixed things.

### DIFF
--- a/R/variables.R
+++ b/R/variables.R
@@ -87,7 +87,7 @@ create_variables <- function(parameters) {
   }
 
   states <- c('S', 'D', 'A', 'U', 'Tr')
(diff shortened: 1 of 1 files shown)


RJSheppard:
@pwinskill that's how I understand it - I just thought it would be better to define the state order in one place (rather than in two), then use it in the initial_states function. Then (in line 322) it was bothering me that the states are in a different order, so I made that consistent. But they aren't necessary and I'm happy for them to go.

giovannic (approved):
Yeah I agree with both of you. Not functionally different, but looks nicer. Thanks!

PR #248, opened 2023-06-28
Vivax asymp infectivity

RJSheppard:
Currently pulling into vivax_parameters to highlight specific changes. This contains switches to change asymptomatic infectivity from an equation into a constant. An alternative could be to set gamma1 to 0 (which would have the same effect), but I think that coopting those parameters for something different could be confusing.


R/human_infection.R
```diff
@@ -347,12 +347,22 @@ schedule_infections <- function(
...
+        timestep,
+        to_infect_asym
+      )
+    } else if (parameters$parasite =="vivax"){ ## P. vivax has a constant asymptomatic infectivity
+      update_infection(
+        variables$state,
+        'A',
+        variables$infectivity,
+        parameters$ca,
+        to_infect
```
pwinskill, line 363: should this be `to_infect_asym`?

tests/testthat/test-vivax.R
```diff
@@ -20,7 +20,62 @@ test_that('Test difference between falciparum and vivax parameter lists', {
   vivax_parameters <- get_parameters(parasite = "vivax")
   
   expect_true(all(names(falciparum_parameters)[!names(falciparum_parameters) %in% names(vivax_parameters)] %in%
-               c("du","rvm","rva","rb","b0","b1","ib0","kb","theta0","theta1","kv","fv0","av","gammav","iv0","fd0","ad","gammad","d1","id0","kd","ub","uv","gamma1","pvm","init_ivm","init_ib","init_iva")))
+                    c("du","rvm","rva","rb","b0","b1","ib0","kb","theta0","theta1","kv","fv0","av","gammav","iv0","fd0","ad","gammad","d1","id0","kd","ub","uv","gamma1","pvm","init_ivm","init_ib","init_iva")))
   expect_true(all(names(vivax_parameters[!names(vivax_parameters) %in% names(falciparum_parameters)]) %in%
                     c("du_max","du_min","ku","au50","b","phi0lm","phi1lm","ic0lm","kclm","d0","ca","init_idm","f","gammal")))
 })
+
```
pwinskill, line 27: In the existing tests there are mocked calls to schedule_infections in tests/testthat/test-infection-integration.R
I think adding to or copying the set up of some of those tests to check the vivax changes to that function would be helpful

RJSheppard: I gave adding a test like this a go - is this what you meant? I think it will make more sense to have this kind of test when subpatent infection is included.

pwinskill (approved):

PR #249, opened 2023-06-28
Vivax pre eryth imm

RJSheppard:
Currently pulling into vivax_parameters to show the key differences. Switch to use the constant b with vivax, rather than using blood_immunity() function.


giovannic (approved):

pwinskill (approved):
:)

PR #250, opened 2023-06-28
Vivax subpatent duration

RJSheppard:
I've created process function to work out subpatent durations (I thought about coopting the existing functions, but they don't line up closely enough and they model conceptually different things). Introduced maternal ID - used in these functions (birth/death and decay). Falciparum probability of detectability functions no longer used, including in rendering (also altered). Tests to check outputs.


R/disease_progression.R
```diff
@@ -84,3 +84,49 @@ update_to_asymptomatic_infection <- function(
...
+
+#' @title Sub-patent duration for vivax model
+#' @description
+#' calculates sub-patent duration based on detectability (anti-parasite) immunity
+#'
+#' @param parameters model parameters
+#' @param variable immunity variables
+#' @param index index of subpatent infected individuals 
+#' @noRd
+subpatent_duration <- function(parameters, variables, index){
```
pwinskill, line 127: Looking at White (2018) supplementary figure 1.
I think this corresponds to Ap (anti-parasite immunity)?
If so I think we should call it that so it references to the paper. I also think you'll call this function in more than one place (the Ap circle is on two flows in that diagram), so good to generalise.

It would actually be good to check if the differently named immunity functions in the Pf and Pv exactly align (for example is vivax anti-parasite immunity == falciparum blood_immunity?). If they do then I think we would want to make sure we don't have identical code with differnt names - ie we would want to use a single function for both Pf and Pv

R/disease_progression.R
```diff
@@ -84,3 +84,49 @@ update_to_asymptomatic_infection <- function(
...
+  variables,
+  parameters
+  ){
+  function(timestep) {
+    ## Get rates for each subpatent individual
+    subpatent_index <- state$get_index_of("U")
+    rates <- subpatent_duration(parameters, variables, index = subpatent_index)
+    ## Get individuals who are going to change
+    to_move <- subpatent_index$sample(1/rates)
+    # tupdate individuals and infectivity
```
pwinskill, line 108: typo

R/disease_progression.R
```diff
@@ -84,3 +84,49 @@ update_to_asymptomatic_infection <- function(
...
+  state,
+  variables,
+  parameters
+  ){
+  function(timestep) {
+    ## Get rates for each subpatent individual
+    subpatent_index <- state$get_index_of("U")
+    rates <- subpatent_duration(parameters, variables, index = subpatent_index)
+    ## Get individuals who are going to change
+    to_move <- subpatent_index$sample(1/rates)
```
pwinskill, line 107: Pls check for consistency in spacing around /, *, == etc

R/disease_progression.R
```diff
@@ -84,3 +84,49 @@ update_to_asymptomatic_infection <- function(
...
+#' @title Sub-patent progression for vivax model
+#' @description
+#' randomly selects individuals to clear sub-patent infections based on impact of 
+#' individual immunity to detectability
+#'
+#' @param state the human state variable
+#' @param variables the available human variables
+#' @param parameters model parameters
+#' @noRd
+create_subpatent_progression_process <- function(
```
pwinskill, line 97: Would be nice for the inputs here to look more like the existing `create_progression_process`, for example I think you can have `from_state` and `to_state` arguments?

R/processes.R
```diff
@@ -169,8 +183,8 @@ create_processes <- function(
     ),
     create_variable_mean_renderer_process(
       renderer,
-      c('ica', 'icm', 'ib', 'id', 'iva', 'ivm'),
-      variables[c('ica', 'icm', 'ib', 'id', 'iva', 'ivm')]
+      c('ica', 'icm', 'ib', 'id', 'iva', 'ivm', if(parameters$parasite == "vivax"){'idm'}),
```
pwinskill, line 186: Would this be more clear (not checked):
```
variable_names <- c('ica', 'icm', 'ib', 'id', 'iva', 'ivm')
if(parameters$parasite == "vivax"){
  variable_names <- c(variable_names, "idm")
}
variable_names, 
variables[variable_names]
```

R/render.R
```diff
@@ -22,14 +22,23 @@ create_prevelance_renderer <- function(
...
+      prob <- probability_of_detection(
+        get_age(birth$get_values(asymptomatic), timestep),
+        immunity$get_values(asymptomatic),
+        parameters
+      )
+      asymptomatic_detected <- bitset_at(asymptomatic, bernoulli_multi_p(prob))
+      
+    } else if (parameters$parasite =="vivax"){
+      # The vivax model defines asymptomatic infections as being detectable by PCR
+      prob <- rep(1,asymptomatic$size())
```
pwinskill, line 38: [nit] spacing

R/render.R
```diff
@@ -22,14 +22,23 @@ create_prevelance_renderer <- function(
...
+    if(parameters$parasite =="falciparum"){
+      prob <- probability_of_detection(
+        get_age(birth$get_values(asymptomatic), timestep),
+        immunity$get_values(asymptomatic),
+        parameters
+      )
+      asymptomatic_detected <- bitset_at(asymptomatic, bernoulli_multi_p(prob))
+      
+    } else if (parameters$parasite =="vivax"){
+      # The vivax model defines asymptomatic infections as being detectable by PCR
```
pwinskill, line 37: But I think prevalence here is (as default) by microscopy? If so then fewer would be detected than via pcr

pwinskill: See Phi_LM in the White et all (2018) SI

tests/testthat/test-vivax.R
```diff
@@ -24,3 +24,53 @@ test_that('Test difference between falciparum and vivax parameter lists', {
...
+  
+  ## Check convergence to min_du at high immunity
+  variables$id <- individual::DoubleVariable$new(rep(1E7,100))
+  variables$idm <- individual::DoubleVariable$new(rep(0,100))
+  expect_equal(object = subpatent_duration(parameters = vivax_parameters, variables = variables, index = index),
+               expected = rep(vivax_parameters$du_min,100),
+               tolerance = 1E-2)
+})
+
+test_that('Test idm rendered creates output', {
```
pwinskill, line 64: We should be able to test this without the model run.
Check out some of Giovanni's examples in tests/testthat/test-render.R

pwinskill (changes requested):
Thanks Richard, great to see all this progress.
A few specific and general comments here from me

R/disease_progression.R
RJSheppard: The section does correspond with anti-parasite immunity, but there are two separate things that are being modelled: duration of sub-patent infection (rate of recovery to S) and probability of patent infection phi_lm. It's on two places in the equation but the modelling is slightly different in each one: the equations mirror each other but have different input parameters; so I'm happy to generalise that. In this pull request I've only the duration, not the probability of patent infection (I actually thought that I already done this, but I think I lost that code trying to discretise the work into smaller chunks! I'll add that back in with the next push).

Screenshot from 2023-06-29 12-12-50

The falciparum and vivax don't align exactly (also, I think we said that the anti-parasite immunity maps onto the detectability immunity (ID), rather than the blood immunity (IB)): the falciparum has what looks like a hill function within another hill function, and only uses acquired ID, not both acquired and maternal ID. It's also conceptually different, because it's the probability that an asymptomatic infection is detectable, so doesn't get used in the same way.

In terms of generalisation, though, there are a number of discrete functions that have the same structure (basically all the hill functions): blood_immunity and clinical_immunity, then probability of detection and severe immunity are both double hill functions. It could be worth generalising those at some point...?

So: I'm going to make a general AP_immunity function that can take parameters from either sub-patent duration or probability of patent infection. Then this can be used in both places.

R/processes.R
RJSheppard: I'll change this, but the section will need to come out of the processes list itself. Maybe that's fine.

R/render.R
RJSheppard: Apologies - this was a typo. PCR should read as "detectable by light-microscopy".

R/render.R
RJSheppard: space added

R/render.R
giovannic: Perhaps this is an opportunity to make the columns more specific, e.g. with `n_detect_microscopy_` for falciparum and `n_detect_LM_` and `n_detect_PCR_` for vivax runs?

We would have to reflect that in the output documentation in `model.R:run_simulation` and maybe the model vignette.

RJSheppard: In the vivax model, the asymptomatic state is defined as being detectable by LM, while sub-patent is defined as being detectable by PCR, so those things are already rendered (unless I've misunderstood). Does that mean we should just remove n-detect for vivax...? Unless we want to add in some sort of probability of LM/PCR detection that isn't just all the individuals in that state...

tests/testthat/test-vivax.R
RJSheppard: I've had a go with this (ripped off Giovanni's tests, really) - see what you think!

pwinskill (approved):
Nice work updating the tests 👍

R/human_infection.R
```diff
@@ -171,15 +190,51 @@ calculate_infections <- function(
   infected
 }
 
-#' @title Calculate clinical infections
+#' @title Calculate patent infections (vivax only)
 #' @description
-#' Sample clinical infections from all infections
+#' Sample patent infections from all infections
```
giovannic, line 195: I don't see doc changes for this, could you commit those?

R/parameters.R
```diff
@@ -334,6 +336,8 @@ get_parameters <- function(overrides = list(), parasite = "falciparum") {
       prevalence_rendering_max_ages = 10 * 365,
       incidence_rendering_min_ages = numeric(0),
       incidence_rendering_max_ages = numeric(0),
+      patent_incidence_rendering_min_ages = numeric(0),
+      patent_incidence_rendering_max_ages = numeric(0),
```
giovannic, line 340: This is not consistent with the documentation. We probably want to change the documentation? Since I imagine changing this would break the run.

R/render.R
```diff
@@ -22,14 +22,24 @@ create_prevelance_renderer <- function(
...
+        get_age(birth$get_values(asymptomatic), timestep),
+        immunity$get_values(asymptomatic),
+        parameters
+      )
+      asymptomatic_detected <- bitset_at(asymptomatic, bernoulli_multi_p(prob))
+      
+    } else if (parameters$parasite == "vivax"){
+      # The vivax model defines asymptomatic infections as being detectable by
+      # light microscopy
+      prob <- rep(1, asymptomatic$size())
```
giovannic, line 39: I kinda love this...

But we should really avoid this prob stuff entirely when doing the vivax model (i.e. p_detect = n_detect)

RJSheppard: For sure - I left this in as something of a cry for help.

I have a couple of wider questions about what's getting rendered and when. I think there are two places where age rendering can currently occur? One in create_age_group_renderer as expected, but then one in create_prevelance_renderer consistent with the prevalence age group. The documentation for create_prevelance_renderer also says it's rendering prevalence of severe malaria, but doesn't look like that's happening in this part of the code...? (The run_simulation documentation also says that n_severe is an output, but I'm unsure of where that happens). I wonder if this function was intended to show severe prevalence, rather than within an age group. I do think that age rendering should probably only happen once. Then for P. vivax, I can report the number in A, D or Tr compartments for a comparable output...

Any thoughts on that would be appreciated!

giovannic: Ah yes! We used to render severe prevalence but that was a mistake, I must've removed the code but not the documentation. Made an issue #262 

> One in create_age_group_renderer

I actually didn't know this was a thing 😆 

But it looks like they are separate features. One is for prevalence statistics (including associated age counts) and the other is for if you want age counts. I'm assuming the latter was introduced so that people could validate their demography configuration or calculate age specific intervention coverage statistics if they wanted to.

> I do think that age rendering should probably only happen once.

I actually disagree. I think that if the user parameterises age counts twice (once with prevalence_rendering and again with age_group_rendering) we should do it twice.

 > Then for P. vivax, I can report the number in A, D or Tr compartments for a comparable output...

I don't follow here sorry...

RJSheppard: I realised yesterday that there are actually three places where age rendering takes place: the prevalence function, the incidence function and the age-rendering function (which means that it's consistent to have age rendering within incidence/prevalence rendering and I now agree with you). But I did also notice that in prevalence and incidence the age outputs are both n_lower_upper, while in the age-rendering the output is n_age_lower_upper. I wonder if the outputs should labelled be the same?

> I don't follow here sorry...

The falciparum model reports n_detect and p_detect in prevalence (LM detectable prevalence, which are clinically detected (Tr or D) or asymptomatic (A)). In vivax the asymptomatic compartment is (by definition) LM detectable, so there is no need to calculate p_detect for vivax. It would still be good to output a comparable metric for vivax, so maybe I'll just output n_detect for both parasites and only p_detect for falciparum. Does that make sense?

RJSheppard: @giovannic @pwinskill 
I've just been checking through some of the previous suggestions and this is what I propose in terms of the n_detect and p_detect e.g.,
> Perhaps this is an opportunity to make the columns more specific, e.g. with n_detect_microscopy_ for falciparum and n_detect_LM_ and n_detect_PCR_ for vivax runs?

Current situation:
- n_detect_ (D + Tr + A)
- p_detect_ (D + Tr + prob(A), where prob due to detectability immunity)

Previously suggested:
- n_detect_LM_ (for Pv, D + Tr + A)
- n_detect_PCR_ (for Pv, D + Tr + A + U)
- n_detect_microscopy (although this seems equivalent to n_detect_LM_, so could condense into one?)

So I suggest:
n_detect_ → n_detect_lm_ (pf and pv: D + Tr + A)
p_detect_ → p_detect_lm_ (pf only: D + Tr + prob(A), where prob due to detectability immunity)
n_detect_pcr_ (new: pf and pv: D + Tr + A + U)

How does that sound? Any thoughts?

RJSheppard: Pete just gave a good explanation for why it's good to have a separate age renderer which is named differently, so please ignore me on that point!

R/render.R
giovannic: Yes, that makes sense! Let's remove n_detect.

R/parameters.R
RJSheppard: Documentation added. Everything seems fine...?

R/human_infection.R
RJSheppard: Documentation added.

R/render.R
RJSheppard: n_detect and p_detect for each parasite species revised now

R/render.R
```diff
@@ -32,38 +35,59 @@ create_prevelance_renderer <- function(
...
-      prob <- rep(1, asymptomatic$size())
-      asymptomatic_detected <- state$get_index_of('A')
+    } else {
+      asymptomatic_detected <- asymptomatic
     }
     
     clinically_detected <- state$get_index_of(c('Tr', 'D'))
-    detected <- clinically_detected$copy()$or(asymptomatic_detected)
-
+    lm_detected <- detected <- clinically_detected$copy()$or(asymptomatic_detected)
```
pwinskill, line 43: This may be something that bugs me an not anyone else, but can we keep these as two seperate assignments please.

pwinskill (commented):
Thanks @RJSheppard 🎉. Just a few bits to follow up on from me.

R/render.R
RJSheppard: Ooops - this was actually a mistake (shouldn't have the middle part). I'll fix that.

R/render.R
```diff
@@ -15,45 +18,76 @@ in_age_range <- function(birth, timestep, lower, upper) {
...
+        
+      } else if (parameters$parasite == "vivax") {
+        
+        # render lm detection (vivax)
+        renderer$render(
+          paste0('n_detect_lm_', lower, '_', upper),
+          in_age$copy()$and(lm_detected)$size(),
+          timestep
+        )
+      }
```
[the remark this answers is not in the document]
RJSheppard: Sure thing - I think I'd been separating them because the calculations were different, but I've since included an extra if statement earlier that does that calculation at an earlier point.

R/render.R
```diff
@@ -104,9 +130,9 @@ incidence_renderer <- function(
 }
 
 create_variable_mean_renderer_process <- function(
-  renderer,
-  names,
-  variables
+    renderer,
```
giovannic, line 133: This is fine now. But adding spaces can cause needless merge conflicts

giovannic (approved):

PR #259, opened 2023-07-18
Feat/metapopulation model

giovannic:


R/mixing.R
```diff
@@ -0,0 +1,72 @@
...
+  export_mixing,
+  import_mixing,
+  p_captured_tt,
+  p_captured,
+  p_success
+  ) {
+  function (timestep) {
+    n_pops <- length(variables)
+    rdt_positive <- vnapply(
+      seq(n_pops),
```
plietar, line 16:
```suggestion
      seq_len(n_pops),
```

`seq` behaves surprisingly when given a size of 0. Admittedly this is unlikely to happen, but safe to just use `seq_len`

R/mixing.R
```diff
@@ -0,0 +1,72 @@
...
+    inf <- vnapply(
+      seq_along(lagged_infectivity),
+      function(i) lagged_infectivity[[i]]$get(timestep - parameters[[i]]$delay_gam)
+    )
+
+    test_and_treat_coeff <- (1 - p_captured_t * rdt_positive * p_success)
+    diag(test_and_treat_coeff) <- 1
+
+    eir <- vapply(
+      seq(n_species),
```
plietar, line 48:
```suggestion
      seq_len(n_species),
```

.gitignore
```diff
@@ -12,3 +12,4 @@ doc
 inst/doc
 /doc/
 /Meta/
+src-i386
```
plietar, line 15: Not sure why this is needed?

plietar (approved):
From my perspective this seems fine, but I don't really have the background context on what is being accomplished.

htopazian:
Thanks @plietar, essentially we are fixing some assumptions in how the metapopulation model mixing matrix is made to allow for different effects on FOIM and EIR. We have done a lot of testing with it, but thanks for providing another set of eyes to the changes!

PR #261, opened 2023-07-24
Vivax hypnozoites

RJSheppard:
The biggest change is to human_infection.R where I've rearranged the order to calculate infections in any state (where hypnozoite batches increases), before subsetting to SAU compartments see where a blood stage infection or relapse can take place. I'm currently rendering n_newly_infected, n_relapses and n_infection, which might be a little overkill, but possibly useful. I'm also rendering average hypnozoite batch number and number of individuals with hypnozoite batches, which can be age-stratified. I'm also rendering hypnozoite decay, although that was mainly to check everything adds up, so could be taken out. Any thoughts on outputs would be useful! Then I have hypnozoite variable creation, decay... I've relabelled some of the parameters to be more consistent with Michael White's paper (and where the code doesn't overlap). I think I might be able to do this with the ca parameter, which would more accurately be the clm (for light-microscopy), rather than ca...

The hypnozoite decay process was a little more finickity than I was expecting. Instead of just applying the standard decay function, I needed something that gave a vector of decaying batches probabilistically, rather than having something that works on a continuous scale (I think that's right - you can't have 0.5 batches in this model).

I also had a go with a test for the relapses.


R/processes.R
```diff
@@ -344,3 +363,25 @@ create_lagged_eir <- function(variables, solvers, parameters) {
...
+create_hypnozoite_decay_process <- function(hypnozoites, gammal, renderer){
+  function(timestep){
+    if(any(hypnozoites$get_values()>0)){
+      hyp_bitset <- hypnozoites$get_index_of(0)$copy()$not(FALSE)
+      hypozoite_decay <- vapply(X = hypnozoites$get_values(index = hyp_bitset),
+                                FUN = function(x){
+                                  rbinom(n = 1, size = x, prob = gammal)
+                                }, numeric(1))
+      hypnozoites$queue_update(values = hypnozoites$get_values(index = hyp_bitset) - hypozoite_decay,
+                               index = hyp_bitset)
```
giovannic, line 383: vnapply in utils.R is exactly for this.

Also, we can save a whole `get_values` here without making the code any uglier:

```suggestion

```

```suggestion
      n_hypnozoites <- hypnozoites$get_values(index = hyp_bitset)
      hypozoite_decay <- vnapply(
        n_hypnozoites,
        function(x) rbinom(n = 1, size = x, prob = gammal)
      )
      hypnozoites$queue_update(
        values = n_hypnozoites - hypozoite_decay,
        index = hyp_bitset
      )
```

R/render.R
```diff
@@ -182,3 +182,38 @@ create_age_group_renderer <- function(
...
+
+create_hypnozoite_renderer_process <- function(
+    renderer,
+    hypnozoites,
+    parameters
+) {
+  function(timestep) {
+    renderer$render(
+      paste0("n_hypnozoites"),
+      sum(hypnozoites$get_values()!=0),
```
giovannic, line 194:
```suggestion
      hypnozoites$size() - hypnozoites$get_size_of(0),
```

R/render.R
```diff
@@ -182,3 +182,38 @@ create_age_group_renderer <- function(
...
+    renderer
+) {
+  function(timestep) {
+    for (i in seq_along(parameters$hypnozoite_rendering_min_ages)) {
+      lower <- parameters$hypnozoite_rendering_min_ages[[i]]
+      upper <- parameters$hypnozoite_rendering_max_ages[[i]]
+      in_age <- in_age_range(birth, timestep, lower, upper)
+      renderer$render(
+        paste0('n_hypnozoites_age_', lower, '_', upper),
+        sum(hypnozoites$get_values(index = in_age$copy()$to_vector())!=0),
```
giovannic, line 213:
```suggestion
        sum(hypnozoites$get_values(index = in_age)!=0),
```

R/human_infection.R
```diff
@@ -183,12 +179,18 @@ calculate_infections <- function(
   }
 
   prob <- b * (1 - prophylaxis) * (1 - vaccine_efficacy)
-  infected <- bitset_at(source_humans, bernoulli_multi_p(prob))
 
+  ## All bitten humans with an infectious bite (incorporating prophylaxis and vaccines)
+  infected <- bitset_at(bitten_humans, bernoulli_multi_p(prob))
```
giovannic, line 184: We should use `bernoulli` here if we're simulating vivax and the probability of infection is constant because it is much faster.

R/human_infection.R
```diff
@@ -130,17 +128,15 @@ calculate_infections <- function(
   renderer,
   timestep
   ) {
-  source_humans <- variables$state$get_index_of(
```
giovannic, line 133: It seems like a mistake to calculate infections for all humans. This would add a lot of computation especially at higher EIRs.

Could we revert this to how it was?

giovannic (changes requested):
Thank you!

R/human_infection.R
RJSheppard: I'm not sure @pwinskill? My understanding is that we need to calculate the superinfections that occur in the D and Tr compartments because those superinfections result in an increase in hypnozoite batch number (although not additional blood stage infection). We could change it back for falciparum at least, including another switch earlier in the code, but vivax might need to be longer...

giovannic: ah, the falciparum model doesn't have superinfection from D or Tr. Also, this actually changes the immunity behaviour of the falciparum model. It probably has to go back to its original behaviour.

If these computations become sufficiently different, it may be cleaner to just have `calculate_vivax_infections` and a `calculate_falciparum_infections` and put a switch in `simulate_infection`

pwinskill: Would it work/be neater to keep the infections flow the same for both parasites (ie with 'S', 'A', 'U'), then add an additional function to capture hypnozoite batch number changes for 'D' and 'Tr'?

R/render.R
RJSheppard: Changed

R/render.R
RJSheppard: Changed

R/human_infection.R
RJSheppard: While the probability of infection (b) is constant, it can still be affected by prophylaxis and prospective vaccines (such that prob is not constant), so bernoulli_multi_p is still appropriate.

R/human_infection.R
RJSheppard: I've rearranged the code to reflect this now. Let me know what you think - I can shove the vivax infections into a new function if you think that would be better.

R/processes.R
```diff
@@ -344,3 +363,28 @@ create_lagged_eir <- function(variables, solvers, parameters) {
...
+#' @title Hypnozoite decay function
+#' @description
+#' calulates the number of decaying hypnozoites in each individual
+#'
+#' @param variable the hypnozoite variable to update
+#' @param rate the hypnozoite decay rate
+#' @noRd
+create_hypnozoite_decay_process <- function(hypnozoites, gammal, renderer){
+  function(timestep){
+    if(hypnozoites$get_size_of(0)>0){
```
pwinskill, line 376: Not sure I understand this line... does this not check if there are any individuals with 0 hypnozoites?

R/human_infection.R
```diff
@@ -181,23 +182,98 @@ calculate_infections <- function(
...
+    total_infected <- newly_infected
+  } 
+  
+  else if(parameters$parasite == "vivax"){
+    
+    SAU_humans <- variables$state$get_index_of(c('S','A','U'))
+    new_bite_infections <- newly_infected$and(SAU_humans)
+    
+    # Render new infections caused by bites
+    renderer$render('n_new_bite_infections', new_bite_infections$size(), timestep)
```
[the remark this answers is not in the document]
RJSheppard: It's not quite the same because an individual can experience a relapse and a new infection at the same time. They aren't necessarily all important outputs (I thought I would include more so we had options), but possibly worth having a conversation about.

R/processes.R
RJSheppard: Oh, good spot. This should be capturing if there are any individuals with hypnozoites. I believe it just needs an "!" to correct that. I'll fix it in the next commit.

R/processes.R
```diff
@@ -344,3 +363,28 @@ create_lagged_eir <- function(variables, solvers, parameters) {
...
+        function(x) rbinom(n = 1, size = x, prob = gammal)
+      )
+      hypnozoites$queue_update(
+        values = n_hypnozoites - hypozoite_decay,
+        index = hyp_bitset
+      )
+      renderer$render('n_decayed', sum(hypozoite_decay), timestep)
+    }
+  }
+}
```
pwinskill, line 390: Is this slightly more readable?

```suggestion
create_hypnozoite_decay_process <- function(hypnozoites, gammal, renderer){
  function(timestep){
    hyp_bitset <- hypnozoites$get_index_of(0)$not(TRUE)
    if(hyp_bitset$size() > 0){      
      n_hypnozoites <- hypnozoites$get_values(index = hyp_bitset)
      hypozoite_decay <- vnapply(
        n_hypnozoites,
        function(x) rbinom(n = 1, size = x, prob = gammal)
      )
      hypnozoites$queue_update(
        values = n_hypnozoites - hypozoite_decay,
        index = hyp_bitset
      )
      renderer$render('n_decayed', sum(hypozoite_decay), timestep)
    }
  }
}
```

giovannic (changes requested):

R/processes.R
RJSheppard: I've added this change in the next commit.

R/processes.R
```diff
@@ -344,3 +363,28 @@ create_lagged_eir <- function(variables, solvers, parameters) {
...
+#' @title Hypnozoite decay function
+#' @description
+#' calulates the number of decaying hypnozoites in each individual
+#'
+#' @param variable the hypnozoite variable to update
+#' @param rate the hypnozoite decay rate
+#' @noRd
+create_hypnozoite_decay_process <- function(hypnozoites, gammal, renderer){
+  function(timestep){
+    if(hypnozoites$get_size_of(!0)>0){
```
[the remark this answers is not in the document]
RJSheppard: I think what we currently have (Pete's suggestion) is working? Happy to get more feedback on this part!

giovannic (approved):
Looks like everything is addressed! Thank you.

Before we merge into dev, we should make a major new version in DESCRIPTION and NEWS.md (since n_detect_lm will probably break people's code).

R/human_infection.R
```diff
@@ -183,21 +184,96 @@ calculate_infections <- function(
...
+      parameters$new_bite_incidence_rendering_min_ages,
+      parameters$new_bite_incidence_rendering_max_ages,
+      timestep
+    )
+
+    ## All bitten humans with an infectious bite (incorporating prophylaxis and vaccines get a new batch of hypnozoites)
+    variables$hypnozoites$queue_update(
+      variables$hypnozoites$get_values(newly_infected) + 1,
+      newly_infected
+    )
```
pwinskill, line 231: We have a competing hazards issues here.
In Michaels model, from your description, they draw all infections, and then split those into either bites or relapses.
In this implementation you draw infections from bites first, and all of those get +1 hypnozoite batches. This will be different from Michaels implementation because a proportion of those new infections from bites would actually be relapses and not get +1 hypnozoite batches. 
Given how sensitive the steady state seems to be regarding hypnozoite dynamics, this could be influential and is therefore something we probably want to align on. In general that would be drawing individuals who get either new bite infection or relapse infection, then splitting those between new bites or relapses based on the relative rates of each.

pwinskill (changes requested):

R/human_infection.R
RJSheppard: I've had a go at this. Let me know if there's anything that I could improve.

RJSheppard: ^Had another go - I think I've corrected the previous errors now.

PR #264, opened 2023-09-15
Fix bug with pev min_wait

giovannic:
This PR addresses an issue where min_wait is apparently ignored. It is easiest to see when you have two mass_pev campaigns close to each other (example below). But would likely affect other combinations of pev interventions (including those using EPI).

```r
library(malariasimulation)

p <- get_parameters(overrides = list(human_population = 10000,
                                     model_seasonality = TRUE,
                                     individual_mosquitoes = FALSE))

p <- set_mass_pev(
  parameters = p,
  profile = rtss_profile,
  timesteps = c(20, 100),
  coverages = c(1, 1),
  min_ages = 0,
  max_ages = 100*365,
  min_wait = 500,
  booster_timestep = 100, # timesteps following initial vaccination
  booster_profile = list(rtss_booster_profile),
  booster_coverage = 1
)

o <- run_simulation(timesteps = 365,
                    parameters = p)

print(sum(o$n_pev_mass_dose_1))
print(o$n_pev_mass_dose_1)
```

The problem was that we were sampling pev populations based on the 3rd dose instead of the 1st. This PR corrects that while maintaining the previous behaviour wrt vaccine efficacy. 

-- Changes --

Fix bug with pev min_wait:
 * Update pev-epi min_wait test to directly check sample population
 * Update pe-epi  min_wait test to check that pev_timestep is being updated on the first dose
 * Copy pev-epi min_wait test to mass pev
 * Update pev and epi processes to calculate targets based on time of first dose
 * Update infection immunity function to calculate pev immunity only after 3rd dose


kellymccain28 (approved):
This is great. Thanks for fixing this Giovanni! 

I tested with multiple mass vaccination campaigns and with a combination of EPI and mass campaigns - all seemed to work as expected.

pwinskill:
I'm being slow and stuggling to follow the logic through all the way.
Can I check that this change still has efficacy starting at the third dose (whist changin min_wait to respect time since first dose)?

pwinskill:
Thanks Giovanni - that sounds sensible to me. I can certaintly see the sitation with multiple mass pev campaigns arising so think we should be robust to that.

giovannic:
@pwinskill I've updated this PR to have an explicit tracking of (non-)efficacious doses. I've updated the existing tests to check the new behaviour and that seems to be enough, but let me know if we should add some more!

pwinskill (approved):
Thanks very much @giovannic. I can follow the logic much more clearly now - much appreciated 🙏

PR #265, opened 2023-09-18
Fix competing hazards for mass and EPI pev:

giovannic:
Fixes are explained in the commit history, please comment if anything is confusing.

Perhaps one point of contention: EPI is run as a process and mass vaccination as an event, so malariasimulation will prioritise EPI over mass vaccination.

For the same reason, the checks in the mass vaccination listener are redundant, but are included defensively. If someone were to change mass vaccination to be a process, or to change the ordering of the vaccination processes to prioritise mass vaccination, the code would maintain the desired behaviour of only enrolling to one vaccination programme.

Let me know if that's to your taste or not.

----Changes----

Fix competing hazards for mass and EPI pev:

 * Add tests for the EPI and mass vaccination processes/listeners
 * Add checks in the process/listener to remove scheduled vaccinations
 * Update the news for the bug fix
 * RcppExports updated automatically

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+# malariasimulation 1.6.1 (wip)
+
+ * Fix bug in competing hazards between mass and EPI vaccines. Where individuals
+   can be enrolled onto both strategies if applied on the same timestep.
+
 # malariasimulation 1.6.0
 
   * Fix MDA bug where undetectable asymptomatics are treated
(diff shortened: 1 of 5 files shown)


kellymccain28 (approved):

pwinskill (approved):
Looks good!
I think totally fine for the EPI process to dominiate the mass event in this case, but thanks for including the defensive tests 🙏

PR #267, opened 2023-11-03
Housing

EllieSherrardSmith:
Richard S and I have added functionality to 'adapt' housing to have a barrier and or killing impact on top of IRS and ITNs


R/vector_control.R
```diff
@@ -65,37 +65,84 @@ prob_bitten <- function(
...
+    spray_on = 1
     rs_comp <- 1 - rs
     ss <- rep(1, n)
     ss[protected_index] <- prob_survives_spraying(
       ks_prime,
       parameters$k0
     )
   } else {
-    phi_indoors <- 0
+    spray_on = 0
```
giovannic, line 76:
```suggestion
    spray_on <- 0
```

R/vector_control.R
```diff
@@ -65,37 +65,84 @@ prob_bitten <- function(
       js_prime,
       parameters$k0
     )
+    spray_on = 1
```
giovannic, line 68:
```suggestion
    spray_on <- 1
```

vignettes/VectorControl_Housing.Rmd
```diff
@@ -0,0 +1,125 @@
...
+```
+  
+### Simulation
+
+Having established a base set of parameters, we can now create a copy of this parameter list and update it to specify that housing has been altered in some way to reduce access to mosquitoes. In the example below, we simulate that 30% of housing is 'improved' in the 2nd year, and that nets are used by a random 50% of the population every three years. We can change the 'repellence' introduced from households multiple times and differently for each mosquito species potentially.
+  
+```{r, fig.align = 'center', out.width='100%'}
+housingparams <- set_housing(
+  simparams,
+  timesteps = housingtimesteps,
```
giovannic, line 92: This needs to be defined somewhere.

You can check that all the automated checks will pass with `devtools::check`

giovannic (changes requested):
Thanks for this!

The vignette doesn't run on the automated build. I've commented in line.

PR #276, opened 2023-12-19
Output render function

RJSheppard:
I had a quick go at creating an age-grouping function. I've only given the option for contiguous age groups, because that seemed simplest, but could try to think of a way to have independent groups?

There are also a few other changes that need a bit more explanation:

I had previously had non-functional code to separately render bite and relapse incidence by age. These would require the correct probabilities, subset for all bitten or all with hypnozoites respectively. Because the probabilities I had were the sum of both, having faffed around converting from probabilities to rates and back to probs, and subsetting, this seemed more hassle than it was worth, and I replaced it with total incidence by age, as found in falciparum. I matched the parameter list the reflect this (and rearranged slightly).

LS prophylaxis parameters have also been replaced with shape and scale parameters.


R/output_parameters.R
```diff
@@ -0,0 +1,40 @@
...
+#' @param severe_incidence age breaks for severe incidence outputs (p.f only); default = NULL
+#' @param prevalence age breaks for clinical prevalence outputs (pcr and lm detectable infections); default = c(730, 3650)
+#' @param hypnozoite_prevalence age breaks for hypnozoite prevalence outputs (p.v only); default = NULL
+#' @export
+#'
+set_epi_outputs <- function(parameters,
+                            age_group = NULL,
+                            incidence = NULL,
+                            patent_incidence = NULL,
+                            clinical_incidence = c(0, 1825),
```
giovannic, line 20:
```suggestion
                            clinical_incidence = NULL,
```

R/output_parameters.R
```diff
@@ -0,0 +1,40 @@
...
+                            severe_incidence = NULL,
+                            prevalence = c(730, 3650),
```
giovannic, line 22:
```suggestion
                            prevalence = NULL,
```

giovannic (changes requested):
I like it!

Small suggestion on the defaults for `set_epi_outputs`

pwinskill (approved):
Like!
...and I learnt a new function: `formals()`

PR #277, opened 2024-01-10
Add asymmetric transmission in mixing:

giovannic:
* Added export_mixing and import_mixing arguments to run_metapop_...
 * Update create_transmission_mixer to mix eir on import and foim on export
 * Fix regression tests

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -131,14 +131,16 @@ run_simulation <- function(
 #' @param correlations a list of correlation parameters for each population
 #' (default: NULL)
 #' @param mixing_tt a vector of time steps for each mixing matrix
-#' @param mixing a list of matrices of mixing coefficients for infectivity.
+#' @param export_mixing a list of matrices of coefficients for exportation of infectivity.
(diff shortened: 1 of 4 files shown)


R/model.R
```diff
@@ -148,32 +150,39 @@ run_metapop_simulation <- function(
...
+
+    for (i in seq_along(mixing)) {
+      if (nrow(mixing[[i]]) != ncol(mixing[[i]])) {
+        stop(sprintf('mixing matrix %d must be square', i))
+      }
+      if (nrow(mixing[[i]]) != length(parameters)) {
+        stop(sprintf("mixing matrix %d's rows must match length of parameters", i))
+      }
+      if (!all(vlapply(seq_along(parameters), function(x) approx_sum(mixing[[i]][x,], 1)))) {
+        warning(sprintf("all of mixing matrix %d's rows must sum to 1", i))
```
htopazian, line 178: we should change to "all of mixing matrix %d's rows do not sum to 1"

PR #278, opened 2024-01-11
Added liver stage drug prophylaxis effects to MDA and SMC code.

RJSheppard:
This is a fairly quick one. I've previously missed adding the liver stage prophylaxis to MDA and SMC code.
I've added an image to show that it's now working (black is no MDA/SMC, red is with blood stage drugs only, green is blood and liver stage drugs).

image


R/mda_processes.R
```diff
@@ -79,6 +79,22 @@ create_mda_listeners <- function(
...
+    # Update liver stage drug effects
+    if(length(parameters$drug_hypnozoite_efficacy)>0){
+
+      successful_hyp_treatments <- bernoulli(
+        length(target),
+        parameters$drug_hypnozoite_efficacy[[drug]]
+      )
+
+      to_clear <- individual::Bitset$new(parameters$human_population)
+      to_clear$insert(target[successful_hyp_treatments])
```
giovannic, line 91: We have a faster implementation of this in utils.R

```suggestion
      to_clear <- sample_bitset(
        target,
        parameters$drug_hypnozoite_efficacy[[drug]]
      )
```

R/pmc.R
```diff
@@ -32,45 +32,62 @@ create_pmc_process <- function(
...
+    # Update liver stage drug effects
+    if(length(parameters$drug_hypnozoite_efficacy)>0){
+
+      successful_hyp_treatments <- bernoulli(
+        length(target),
+        parameters$drug_hypnozoite_efficacy[[drug]]
+      )
+
+      to_clear <- individual::Bitset$new(parameters$human_population)
+      to_clear$insert(target[successful_hyp_treatments])
```
giovannic, line 85:
```suggestion
      to_clear <- sample_bitset(
        target,
        parameters$drug_hypnozoite_efficacy[[drug]]
      )
```

giovannic (changes requested):

R/mda_processes.R
RJSheppard: Oh cool - I had basically just copied the blood stage prophylaxis code in these sections. I'm guessing that these can also be changed to use the sample_bitset function. e.g., in mda_processes:
image

I'll make the same change to these parts, assuming there isn't a reason this code needs to be different.

giovannic:
Oh good shout. That would be great!

R/mda_processes.R
RJSheppard: I've made the change, but have had to leave in the creation of a new bitset (`target_bit`) because the current target is only a vector. I think it's still more efficient (at least in terms of space).

R/human_infection.R
```diff
@@ -442,7 +445,7 @@ calculate_treated <- function(
   # Update liver stage drug effects
   if(length(parameters$drug_hypnozoite_efficacy)>0){
 
-    ## Only ls stage drugs have hypnozoite efficacy, so liver stage drug variable will be consistent in prophylaxis
+    ## When ls stage drug efficacy is NA, this does not results in a success
```
pwinskill, line 448:
```suggestion
    ## When ls stage drug efficacy is NA, this does not result in a success
```

pwinskill (approved):
Looks good 👍

giovannic (approved):
Looks great!

PR #282, opened 2024-02-15
Added check for undeveloped antimalarial resistance features (SPC/LPF…

tbreweric:
Following the suggestion in the antimalarial_resistance/etf pull request, I have added a check in the `set_antimalarial_resistance()` function to error when users attempt to give non-zero values to parameters relating to undeveloped features.

I have also added a test to the test-antimalarial-resistance.R file to ensure the check works as anticipated.

The documentation entries for the parameters in the `get_parameters()` (parameters.R) documentation have been modified to warn users that the parameters represent features under development.

I stopped short of adding a note in the vignette, but can add this if it is thought to be a good idea!

### DIFF
--- a/R/antimalarial_resistance.R
+++ b/R/antimalarial_resistance.R
@@ -26,6 +26,14 @@ set_antimalarial_resistance <- function(parameters,
                                         reinfection_prob,
                                         slow_parasite_clearance_time) {
   
+  if(any(partner_drug_resistance > 0,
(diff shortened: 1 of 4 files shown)


pwinskill (approved):
Thanks Tom, that's a good safety net for now 😎.
Well done for adding the test on the error - that will ensure you don't forget to turn it off when those features are added.

giovannic (approved):
👍

PR #286, opened 2024-03-22
Allow simulation to be restored with new interventions.

plietar:
This uses a few of individual's new features that make restoring more flexible. It also fixes a bug when restoring the mosquitto solvers, by correctly restoring the timestep, which is needed to model seasonality.

The intervention events use the new `restore = FALSE` flag to make sure their schedule can be modified when resuming. Instead of having the events re-schedule themselves everytime they fire, we setup the entire schedule upfront when initialising the simulation.

It adds end-to-end testing of this feature, across a range of scenarios. For each scenario, the outcomes of the simulation with and without restoring are compared and we make sure they are equivalent.


tests/testthat/test-resume.R
```diff
@@ -1,41 +1,147 @@
-test_that('Simulation can be resumed', {
-  initial_timesteps <- 50
-  total_timesteps <- 100
-
-  parameters <- get_parameters()
+#' Test simulation saving and restoring for a given parameter set.
+#'
+#' This function runs the simulation twice. A first, continous and uninterrupted
```
giovannic, line 3:
```suggestion
#' This function runs the simulation twice. A first, continuous and uninterrupted
```

tests/testthat/test-resume.R
```diff
@@ -1,41 +1,147 @@
...
+#' run of the simulation is used as a reference. The second run is split into
+#' two phases. Between the two phases, the simulation state is saved and
+#' restored. Optionally, the initial warmup phase can use a different set of
+#' parameters, by specifying a value for warmup_parameters.
+test_resume <- function(
```
giovannic, line 8: Really handy

tests/testthat/test-resume.R
```diff
@@ -1,41 +1,147 @@
...
+
+  # The first dose, at time step 25, happens during the warmup and is not
+  # returned by test_resume, hence why we don't see it in the data. Follow-up
+  # doses do show up, event though they we scheduled during warmup.
+  expect_equal(data[data$n_pev_mass_dose_1 > 0, "timestep"], c(100))
+  expect_equal(data[data$n_pev_mass_dose_2 > 0, "timestep"], c(70, 145))
+  expect_equal(data[data$n_pev_mass_dose_3 > 0, "timestep"], c(115, 190))
+})
+
+test_that("TBV intervention can be added when resuming", {
```
giovannic, line 92: This test seems different from the others, should we name it differently? or remove the TBV parameterisation from the warmup?

```suggestion
test_that("TBV intervention parameters can changed when resuming", {
```

giovannic (changes requested):
Lovely!

Perhaps a test case for setting correlation? Since it seems an important use case that you spent some time resolving.

plietar:
> Perhaps a test case for setting correlation? Since it seems an important use case that you spent some time resolving.

Setting correlation doesn't work in this PR yet. I'll have a follow up soon that adds it with some test cases.

tests/testthat/test-resume.R
plietar: I've tweaked the layout of the test a little bit which should make it easier to spot the two parts of it.

giovannic (approved):
Please merge in #293 first. Thanks!

PR #288, opened 2024-04-04
Update description of the dt parameter.

plietar:
The dt parameter was described as "the delay to move from state Tr to Ph", but the Ph (prophylaxis) state does not exist in the model. Instead the dt parameter is used to transition from Tr to S, in `R/processes.R`.

Looking at the commit history, there seems to have been a bit of back and forth around adding and removing the Tr and Ph states. The comment in question was introduced in mrc-ide/malariasimulation#30, after the prophylaxis state had already been removed and even at the time the parameter was already used as the transition rate from Tr to S.

### DIFF
--- a/R/parameters.R
+++ b/R/parameters.R
@@ -11,7 +11,7 @@
 #' fixed state transitions:
 #'
 #' * dd - the delay for humans to move from state D to A; default = 5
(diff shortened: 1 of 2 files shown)


pwinskill (approved):
Thanks Paul. Good spot. You're right that this has been an area of the model that has changed in the past.

PR #289, opened 2024-04-05
Antimalarial Resistance: Slow Parasite Clearance

tbreweric:
**Integrating Slow Parasite Clearance into the Antimalarial Resistance Model**

_Introduction_
In this PR, I have added slow parasite clearance (SPC) as an outcome of malaria infections that are resistant to the artemisinin component of the artemisinin combination therapy (ACT) an individual is treated with. SPC is the phenomenon in which resistant malaria parasites remain in the blood beyond the usual 3 day period taken for the artemisinin component to remove malaria parasites from the patients' blood stream. The parasites are eventually removed by the ACT and the patient successfully treated. 

_Explanation of Slow Parasite Clearance Code Changes_
We have captured SPC by increasing the average time individuals with a clinical infection treated with an ACT spend in the treated compartment. This has involved making changes to eight core model functions, which I will describe below:

1. `set_antimalarial_resistance()`: `slow_parasite_clearance_probability `(formerly `slow_parasite_clearance_prob`) has been removed from the test for undeveloped model features. The names of the resistance parameters have also been amended to make them clearer (e.g. `reinfection_prob` to `reinfection_during_prophylaxis_probability`.
2. `get_parameters()`: The documentation and entries for the resistance parameters have been to updated to reflect the changes to some of their names.
3. `create_variables()`: If parameters$antimalarial_resistance == TRUE `dt` is set up as a variable, initialised with `parameters$dt` for all individuals, and appended it to the variables list. A line has also been added to the documentation that explains the new dt variable.
4. `create_progression_process()`: The original `create_progression_process()` function was only used when the rate at which individuals transition from the `from_state` to the `to_state` was a single parameter value. When `parameters$antimalarial_resistance == TRUE`, the rate at which individuals move from the treated compartment to the susceptible compartment becomes a variable. When the rate becomes a variable an additional step is required to subset the variable values corresponding to the individuals in the `from_state`. The `create_progression_process()` has been amended to first retrieve and store the indices of all individuals in `from_state`. If `length(rate) > 1`, telling the function that the rate is a variable rather than a parameter, the rate is overwritten with only the values of the individuals in the `from_state` in the current timestep. These individual-specific rates are then used to determine which individuals will move from the `from_state` to the `to_state` in the current timestep.
5. `create_processes()`:  I’ve replaced the create_progression_process() that governs the disease state transition from Tr to S. The updated version stores the parameters$dt as the dt_input and overwrites this with variables$dt if antimalarial_resistance == TRUE. The input for the create_progression_process() is then dt_input.
6. `calculate_treated()`: The major change in this PR is within the `calculate_treated()` function. The updated version can simulate both ETF and SPC as outcomes of antimalarial resistance. In the updated version, there are four possible outcomes for individuals that receive treatment:
     - Treatment failure due to drug efficacy (`variables$state `updated to D)
     - Treatment failure due to early treatment failure (`variables$state `updated to D)
     - Successful treatment with slow parasite clearance (`variables$state `updated to Tr, `variables$dt` updated to `parameters$slow_parasite_clearance_time`)
     - Successful treatment without slow parasite clearance (`variables$state `updated to Tr, `variables$dt` updated to `parameters$dt`)

The order of events has been updated. Once the drugs have been assigned to each individual in `seek_treatment`,  the drug efficacy test is performed, with individuals who pass the drug efficacy step stored in the Bitset `effectively_treated` and the drugs vector updated to filter out the drugs of individuals who failed treatment due to drug efficacy. 

If `parameters$antimalarial_resistance == TRUE`, the `get_antimalarial_resistance_parameters()` function is called to retrieve the resistance parameters associated with the drug administered to each individual in `effectively_treated` the current timestep (stored in object called `resistance_parameters`). Each individuals' probability of experiencing early treatment failure is calculated by multiplying the `artemisinin_resistance_proportion` and `early_treatment_failure_probability` corresponding to the drug they were administered and the current timestep. These probabilities are used by `bernoulli_multi_p()` to determine which individuals are successfully treated (avoid treatment failure due to drug efficacy or drug resistance) and their indices are stored in the Bitset `successfully_treated`. The `drugs` and `dt_slow_parasite_clearance` vector (stored in `resistance_parameters`) are subsetted to include only those individuals who have been successfully treated.

For each of the `successfully_treated` individuals, a final test is performed to see whether they will experience slow parasite clearance (again via `bernoulli_multi_p()`). Additional Bitsets are created to store individuals who will (`slow_parasite_clearance_individuals`) and will not experience slow parasite clearance (`non_slow_parasite_clearance_individuals`).

For the `queue_update` operations, the `successfully_treated` Bitset is used to queue updates to the state, infectivity, drug, and `drug_time` variables. For `non_slow_parasite_clearance` individuals, the `dt` variable is updated with the default/non-resistant `parameters$dt`. For the `slow_parasite_clearance_individuals`, the `dt` variable is updated to the `parameters$dt_slow_parasite_clearance` value for the drug the individual was administered (from `resistance_parameters$dt_slow_parasite_clearance`).

When `parameters$antimalarial_resistance == TRUE`, additional outputs are rendered to capture the number of people who received treatment (`n_treated`), the number of individuals who failed treatment due to drug efficacy (`n_drug_efficacy_failures`) or early treatment failure (`n_early_treatment_failures`), the number of people who experienced slow parasite clearance (`n_slow_parasite_clearance`), and the number of individuals who were successfully treated (`n_successfully_treated`). Note that `n_successfully_treated` includes individuals with slow parasite clearance as, although they will spend longer in the Tr compartment, they do transition to Tr and eventually S. 

If `parameters$antimalarial_resistance == FALSE`, the drug resistance steps are skipped, the `effectively_treated` Bitset is copied into `successfully_treated`, and only the `queue_update` steps for `state`, `infectivity`, `drug`, and `drug_time` are performed.

8. `reset_target()`: When `parameters$antimalarial_resistance == TRUE`, a new variable `dt` is created. The `reset_target()` function has been amended to accept the `parameters` list as an input and reset the `dt` variable to `parameters$dt` when `!is.null(variables$dt)`.
9. `run_simulation()`: I have added entries to the `run_simulation()` documentation to describe the new model outputs associated with antimalarial resistance. When `parameters$antimalarial_resistance == TRUE`, the function will now render the following outputs:
        - `n_treated`: number of individuals that were given an ACT in the current timestep
        - `n_drug_efficacy_failures`: number of individuals who failed treatment due to the drug efficacy in the current timestep
        - `n_early_treatment_failures`: number of individuals who failed treatment due to early treatment failure in the current timestep
        - `n_slow_parasite_clearance`: number of individuals who will experience slow parasite clearance in the current timestep
        - `n_successfully_treated`: number of individuals who were successfully treated in the current timestep. Note that this includes individuals with SPC (`n_slow_parasite_clearance`) as they progress to the Tr compartment and will be successfully treated after an increased residency in the compartment. To calculate the number of non-resistant successfully treated individuals the user can simply do `n_successfully_treated `- `n_slow_parasite_clearance`. Furthermore, of the new parameters, `n_drug_efficacy_failures`, `n_early_treatment_failures`, and `n_successfully_treated` sum to `n_treated`.

10. `Antimalarial_resistance.Rmd`: I have amended the antimalarial resistance vignette to capture the changes in the names of the resistance parameter inputs for `set_antimalarial_resistance()` and references to them in the text. I have also added an example of simulations with multiple resistance outcomes - comparing a no-intervention baseline scenario with treatment-only, ETF-only, SPC-only, and combined ETF and SPC simulations.

_Additional Changes_
- Additional tests have been added to test-antimalarial_resistance.R and test-infection-integration.R and, where necessary, tests have been fixed where the changes described above have broken them.
- Changed all instances of `simparams` to `parameters` in unit tests for consistency with other tests.


R/human_infection.R
```diff
@@ -295,50 +295,81 @@ calculate_treated <- function(
...
+    
+    #+++ EARLY TREATMENT FAILURE +++#
+    #+++++++++++++++++++++++++++++++#
+    early_treatment_failure_probability <- resistance_parameters$artemisinin_resistance_proportion * resistance_parameters$early_treatment_failure_probability
+    successfully_treated_indices <- bernoulli_multi_p(p = 1 - early_treatment_failure_probability)
+    successfully_treated <- bitset_at(effectively_treated, successfully_treated_indices)
+    n_early_treatment_failure <- effectively_treated$size() - successfully_treated$size()
+    renderer$render('n_early_treatment_failure', n_early_treatment_failure, timestep)
+    drugs <- drugs[successfully_treated_indices]
+    resistance_parameters$dt_slow_parasite_clearance <- resistance_parameters$dt_slow_parasite_clearance[successfully_treated_indices]
```
giovannic, line 323: Are we sure we want to reassign this back to `resistance_parameters`?

```suggestion
    dt_slow_parasite_clearance <- resistance_parameters$dt_slow_parasite_clearance[successfully_treated_indices]
```

R/processes.R
```diff
@@ -104,12 +104,32 @@ create_processes <- function(
...
+  # =======================
+  # Add an a new process which governs the transition from Tr to S when
+  # antimalarial resistance is simulated. The rate of transition switches
+  # from a parameter to a variable when antimalarial resistance == TRUE.
+  
+  # Assign the dt input to a separate object with the default single parameter value:
+  dt_input <- parameters$dt
+  
+  # If antimalarial resistance is switched on, assign dt variable values to the 
+  if(parameters$antimalarial_resistance == TRUE) {
```
giovannic, line 121:
```suggestion
  if(parameters$antimalarial_resistance) {
```

R/human_infection.R
```diff
@@ -295,50 +295,81 @@ calculate_treated <- function(
...
+    #+++ SLOW PARASITE CLEARANCE +++#
+    #+++++++++++++++++++++++++++++++#
+    slow_parasite_clearance_probability <- resistance_parameters$artemisinin_resistance_proportion[successfully_treated_indices] *
+      resistance_parameters$slow_parasite_clearance_probability[successfully_treated_indices]
+    slow_parasite_clearance_indices <- bernoulli_multi_p(p = slow_parasite_clearance_probability)
+    slow_parasite_clearance_individuals <- bitset_at(successfully_treated, slow_parasite_clearance_indices)
+    renderer$render('n_slow_parasite_clearance', slow_parasite_clearance_individuals$size(), timestep)
+    non_slow_parasite_clearance_individuals <- successfully_treated$copy()$set_difference(slow_parasite_clearance_individuals)
+    renderer$render('n_successfully_treated', successfully_treated$size(), timestep)
+    resistance_parameters$dt_slow_parasite_clearance <- resistance_parameters$dt_slow_parasite_clearance[slow_parasite_clearance_indices]
```
giovannic, line 334:
```suggestion
    dt_slow_parasite_clearance <- resistance_parameters$dt_slow_parasite_clearance[slow_parasite_clearance_indices]
```

R/human_infection.R
```diff
@@ -295,50 +295,81 @@ calculate_treated <- function(
...
       timestep,
       successfully_treated
     )
+    if(parameters$antimalarial_resistance) {
+      variables$dt$queue_update(
+        parameters$dt,
+        non_slow_parasite_clearance_individuals
+      )
+      variables$dt$queue_update(
+        resistance_parameters$dt_slow_parasite_clearance,
```
giovannic, line 363:
```suggestion
        dt_slow_parasite_clearance,
```

R/variables.R
```diff
@@ -226,6 +229,15 @@ create_variables <- function(parameters) {
     net_time = net_time,
     spray_time = spray_time
   )
+  
+  # Add variables for antimalarial resistance state residency times (dt)
+  if(parameters$antimalarial_resistance == TRUE) {
```
giovannic, line 234:
```suggestion
  if(parameters$antimalarial_resistance) {
```

tests/testthat/test-infection-integration.R
```diff
@@ -855,62 +893,62 @@ test_that('calculate_treated() returns an empty Bitset when the parameter list c
...
+                                                      slow_parasite_clearance_time = 20)
+            
+            clinical_infections <- individual::Bitset$new(100)
+            clinical_infections$insert(sample.int(n = 100, size = round(runif(n = 1, min = 10, max = 100)), replace = FALSE))
+            timestep <- 5
+            events <- create_events(parameters)
+            variables <- create_variables(parameters = parameters)
+            renderer <- individual::Render$new(timesteps = 10)
+            
+            treated <- calculate_treated(variables = variables,
```
giovannic, line 942: Could we fix the indentation here to match the other tests? It could be a tab vs spaces thing

giovannic (changes requested):
This is incredible! I've requested a few stylistic changes.

Thank you so much

R/disease_progression.R
```diff
@@ -19,15 +19,27 @@ update_infection <- function(
...
+) {
   function(timestep) {
-    to_move <- state$get_index_of(from_state)$sample(1/rate)
+    
+    # Retrieve the indices of all individuals in the to_move state:
+    index <- state$get_index_of(from_state)
+    
+    # If the length of rate is greater than 1 (when it's a variable):
+    if (length(rate) > 1) {
+      rate <- rate$get_values(index$to_vector())
```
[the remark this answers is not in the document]
tbreweric: Much neater - made the change as recommended!

R/human_infection.R
tbreweric: I've implemented this change as recommended here, but further down write over the new `dt_slow_parasite_clearance` as otherwise I don't indices from the bernoulli_multi_p() calls will correspond to the values they should.

R/processes.R
tbreweric: Change implemented as recommended!

R/human_infection.R
tbreweric: I have opted instead for:

dt_slow_parasite_clearance <- dt_slow_parasite_clearance[slow_parasite_clearance_indices] to ensure the indices match the way (I think) they should!

R/human_infection.R
tbreweric: Change implemented as suggested!

R/variables.R
tbreweric: Change implemented as suggested!

tests/testthat/test-infection-integration.R
tbreweric: Good spot - I've fixed the indentation so that it is consistent between all tests

vignettes/Antimalarial_Resistance.Rmd
```diff
@@ -259,5 +258,157 @@ text(x = resistance_update_timesteps[2:5] + 30, y = 0.6, labels = paste0("Art. R
...
+  # Print the number of columns in the i-th simulation outputs dataframe:                            
+  print(ncol(simulation_temp))
+}
+
+```
+
+### Data Wranging
+
+
+
```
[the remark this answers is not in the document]
tbreweric: Good spot - I've removed this redundant section!

tbreweric:
Thanks @giovannic! I've gone through and made all of the changes - just re-running the checks now but shall commit/push the changes shortly!

R/disease_progression.R
```diff
@@ -19,15 +19,27 @@ update_infection <- function(
...
+    # Retrieve the indices of all individuals in the to_move state:
+    index <- state$get_index_of(from_state)
+    
+    # If the length of rate is greater than 1 (when it's a variable):
+    if (length(rate) > 1) {
+      rate <- rate$get_values(index)
+    }
+    
+    # Sample the individuals to be moved into a new Bitset using the transition rate(s):
+    to_move <- index$sample(1/rate)
```
pwinskill, line 40: I've been trying to trace back through the code to see if this is a rate or a probability and it caused me some confusion.

As used here the parameter `rate` is actually an average duration so is misnamed.
That duration gets converted to a rate in `1/rate` and provided as an argument to `$sample`. Would this be possible to clarify in the malariasimulation code @tbreweric?

@giovannic going deeper into this rabbit hole....I looked at the sample function in th individual package [here](https://github.com/mrc-ide/individual/blob/4bef32e1f2bb12796046467c88343902b5efc08c/inst/include/IterableBitset.h#L633-L653). That details an argument called `rate` which is actually a probability.

We should be careful here as in the code here `1/rate` is actually a rate and not a probability. The approximation only holds if the rate is small. For most cases this is probably fine, but I think at least the documentation and naming should be clear and precise. Sorry if this seems pedantic!

pwinskill (changes requested):
Thanks @tbreweric - this looks really nice! 🥳
I've left one key point regarding rates/probs
Interested to know what others think about the very clear but long variable names.

tbreweric:
> Thanks @tbreweric - this looks really nice! 🥳 I've left one key point regarding rates/probs Interested to know what others think about the very clear but long variable names.

I wondered if I'd gone too far the other way after we talked about naming clarity / length as we'd discussed it during one of our early code reviews. Happy to change them!

R/disease_progression.R
giovannic: Ah that's a good point on the individual side! I've mad an [issue](https://github.com/mrc-ide/individual/issues/196) to rename those variable names and docs to make it clearer. 

[Richard's utilities](https://github.com/mrc-ide/malariasimulation/blob/vivax_main/R/utils.R#L68-L74) (which assume an exponential process) may help make these conversions clearer on the malariasimulation side, if that's what you want to do.

R/disease_progression.R
```diff
@@ -19,15 +19,27 @@ update_infection <- function(
...
+    new_infectivity
+) {
   function(timestep) {
-    to_move <- state$get_index_of(from_state)$sample(1/rate)
+    
+    # Retrieve the indices of all individuals in the to_move state:
+    index <- state$get_index_of(from_state)
+    
+    # If the length of rate is greater than 1 (when it's a variable):
+    if (length(rate) > 1) {
```
plietar, line 35:
```suggestion
    if (inherits(rate , "DoubleVariable")) {
```

This should be equivalent, but it demonstrates the intent more clearly IMO.

R/disease_progression.R
tbreweric: Not sure how best to proceed here as I think there are two options:
1. Change the name of the input in `create_progression_process()` to `average_residence_time `(or something better) so that you are then more obviously calculating the rate by taking the inverse of this, and then feed `$sample` a `rate`. The other instances where `create_progression_process()` is called do not use the arg name so wouldnt need to be amended I don't think. 

2. Change the inputs for `rate` in the `create_progression_process()` calls in `create_processes()` to `1/parameters$dx` so that we are feeding the function a rate.

With each of these, I'm wondering whether this PR is the place to make the changes or whether we make it the subject of it's own for clarity? What do you all think?

R/disease_progression.R
tbreweric: I've never used this approach before but I agree it's clearer about what we're checking for - will make the change! Thanks @plietar!

R/disease_progression.R
tbreweric: @pwinskill @giovannic Is there a preference for what to do here? My instinct is to address this is a separate, small PR but happy to go with whichever you guys think is best!

tbreweric: I have made an issue to address this (#290). I have also made a branch (bug/progression_process_rates) with a proposed fix (changing the rate inputs in the `create_progression_processes()` to `1/parameters$dX` and changing the `sample()` input to `rate `from `1/rate`.

I think if we merge this branch first (if everyone is satisfied with the changes), then I'll amend the branch Ive made to make it compatible with this one and then submit a PR for that. That should give us something clearer in the short term before the proposed changes to the `sample()` function in individual?

giovannic: Hi sorry for the delay on this.

Thank you for making a separate issue for this. I agree that it is out of scope for this PR, which should focus on SPC.

Let's carry on this conversation in #290

giovannic (approved):
My requested changes have been addressed. Thanks!

pwinskill (approved):

PR #291, opened 2024-04-22
Clarify some intervention parametrization documentation.

plietar:
The `set_tbv` function was located at the bottom of `pev_parameters.R`, which was confusing and unintuitive. I believe this was a remnant of a time when that file was named `vaccine_parameters.R`. It is now moved to its own file, named `tbv_parameters.R`. This matches the general naming convention used by all other interventions.

The ages parameter of `set_tbv` was previously described as being "per-round", which might hint at it being, eg. a list of age ranges with one entry per round. This did not match the implementation, which uses the same whole `tbv_ages` parameter for every round.

I don't how useful being able to pass in a non-contiguous set of ages is. In the future it might be clearer to change the TBV intervention to use `min_ages` and `max_ages`, the same way other mass-event interventions (MDA, SMC and mass-PEV) are implementated. This will make it easy to actually have varying age ranges for each round, if desired.

This also adds a check in `set_tbv` for the length of coverages to ensure it matches the timesteps, the same way it is done in other interventions.

The documentation for the `timesteps` and `coverages` parameters of the `set_pmc` method made mentions of rounds, implying that the intervention only takes place at the given timesteps, when it is in fact a perennial age-based intervention that may happen at any timestep. The timesteps parameter is only used to vary the coverage over time.

Finally, there were a pair of lost words in the documentation of `set_tbv` and `set_mda` from successive code refactors, which have now been removed.


pwinskill (approved):
Thanks Paul that is heplful. I think some of the PMC documentation probably got copied across from SMC, but you are right that the implementation is different. 

I agree that ideally at somepoint `tbv` specification would macth other interventions. I'll make an issue and when someone is next working on it will ask if they can update.

PR #292, opened 2024-04-23
Moved mortality to the end of processes to resolve order of competing…

RJSheppard:
I have moved the mortality process to the end and created a couple of plots to show the simulation differences:

I ran 10 simulations for ten years over EIR =1,5,10,50,100 (human_population = 100) for each version.

Average with 95% CrI
[competing_hazards_mortality_process_position.pdf](https://github.com/mrc-ide/malariasimulation/files/15073383/competing_hazards_mortality_process_position.pdf)

Timesteps with smoothed lines
[competing_hazards_mortality_process_position_timesteps.pdf](https://github.com/mrc-ide/malariasimulation/files/15073384/competing_hazards_mortality_process_position_timesteps.pdf)

Prevalence may be a little higher with mortality at the end due to more deaths leading to more susceptibles without immunity leading to more infections? but there's not much in it.

Any thoughts, please let me know!


pwinskill (changes requested):
Thanks Richard. This ordering makes much more sense to me. It appears to be having very little impact on transmisison as we hoped. 

Not sure why the blank lines are showing up as differnces in this PR.

One suggestion would be to run a quick check on the resulting age distributions at the end of the runs, before and after this change. That would be the other obvious place mortality processes would show up potential differences?

tbreweric (approved):
Looks OK to me! The plots look reassuring too! Can't think of any additional checks beyond @pwinskill's suggestion!

tbreweric:
> Thanks Richard. This ordering makes much more sense to me. It appears to be having very little impact on transmisison as we hoped.
> 
> Not sure why the blank lines are showing up as differnces in this PR.
> 
> One suggestion would be to run a quick check on the resulting age distributions at the end of the runs, before and after this change. That would be the other obvious place mortality processes would show up potential differences?

I think it's because Richard had deleted spaces on these lines?

pwinskill (approved):
🎉

giovannic (approved):
🙏

RJSheppard:
Thanks for approving, for fun I'll also just share the difference in immunities.
Again, not much of a difference.
[competing_hazards_mortality_process_position_end_age_structure.pdf](https://github.com/mrc-ide/malariasimulation/files/15094947/competing_hazards_mortality_process_position_end_age_structure.pdf)

PR #293, opened 2024-04-23
Allow correlations to be extended when resuming.

plietar:
When resuming a simulation, it is possible to add new intervention. The correlation of the new intervention may need configuring, both relative to itself and to existing interventions.

The correlation parameters work by sampling, at the start of the simulation, relative weights for each individual and intervention. The weight are later used to select which individuals are targeted by the a given intervention. There weights are drawn from a multivariate normal distribution, whose variance-covariance matrix depends on the configured correlation.

When adding interventions, new corresponding columns need to be added to the weights matrix. A fresh mvnorm distribution cannot be used, since it would override the already drawn weigths for the existing interventions. The solution instead is to use a conditional mvnorm distribution on the new columns, using the existing values as the conditions. This yields new weigths which follow the determined variance-covariance matrix while preserving the existing values.

### DIFF
--- a/R/correlation.R
+++ b/R/correlation.R
@@ -9,7 +9,26 @@ INTS <- c(
 )
 
 #' Class: Correlation parameters
-#' Describes an event in the simulation
+#'
+#' This class implements functionality that allows interventions to be
+#' correlated, positively or negatively. By default, interventions are applied
+#' independently and an individual's probability of receiving two interventions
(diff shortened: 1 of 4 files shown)


plietar:
This should be the final missing piece for the restore functionality to be properly useable :partying_face:

giovannic (approved):
Great docs, thanks!

PR #296, opened 2024-05-01
calculate_eir function correction

RJSheppard:
The function that calculates the initial EIR doesn't appear to be quite right. I don't think it needs to be multiplied by psi. When these values are drawn in lines 135-140 of biting_process.R they get multiplied by psi in line 144. That means we're effectively multiplying these values by psi^2. I think we can correct this by adjusting the calculation thusly.

The two tests at the end would then need adjusting (one way or another) to reflect this change - which could probably use some improving.

I've drawn a couple of plots to show the difference this is trying to correct and that it doesn't change too much:
[eir_calculation_comp_timesteps.pdf](https://github.com/mrc-ide/malariasimulation/files/15178558/eir_calculation_comp_timesteps.pdf)
[eir_calculation_comp.pdf](https://github.com/mrc-ide/malariasimulation/files/15178561/eir_calculation_comp.pdf)

### DIFF
--- a/tests/testthat/test-biology.R
+++ b/tests/testthat/test-biology.R
@@ -16,17 +16,11 @@ test_that('total_M and EIR functions are consistent with equilibrium EIR', {
     vector_models <- parameterise_mosquito_models(parameters, 1)
     solvers <- parameterise_solvers(vector_models, parameters)
     estimated_eir <- calculate_eir(1, solvers, variables, parameters, 0)
-    age <- get_age(variables$birth$get_values(), 0)
-    psi <- unique_biting_rate(age, parameters)
-    omega <- mean(psi)
-    mean(estimated_eir) / omega / population * 365
+    mean(estimated_eir) / population * 365
   })
 
   expect_equal(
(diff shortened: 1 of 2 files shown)


tests/testthat/test-biology.R
```diff
@@ -23,7 +23,10 @@ test_that('total_M and EIR functions are consistent with equilibrium EIR', {
   })
 
   expect_equal(
-    actual_EIR,
+    actual_EIR * 0.766,
```
giovannic, line 26: It seems cleaner to remove omega from above. Pre-calculated values make tests quite brittle.

giovannic (changes requested):

tests/testthat/test-biology.R
RJSheppard: Sure thing! Have done :)

pwinskill (approved):
🙏
Thanks Richard. That little step change had puzzled me previously. Good to squash it.

giovannic (approved):

PR #298, opened 2024-05-02
lm pcr prevalence explicit

RJSheppard:
malariasimulation currently only outputs lm prevalence, but it may be useful to also output pcr prevalence. I've added this functionality in.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -69,15 +69,13 @@ LazyData: true
 Remotes:
   mrc-ide/malariaEquilibrium,
   mrc-ide/individual
-Additional_repositories:
-  https://mrc-ide.r-universe.dev
 Imports:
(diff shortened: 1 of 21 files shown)


giovannic (approved):
I approve! As long as `n_detect_pcr` including all of U is something that users would expect. @pwinskill ?

pwinskill:
> I approve! As long as `n_detect_pcr` including all of U is something that users would expect. @pwinskill ?

Good question Giovanni!
@RJSheppard - perhaps you could chat with Lucy at somepoint just to check this. I think she will have a much better idea of what is reasonable to assume here?

RJSheppard:
Just checked with Lucy and she says that's all good!

pwinskill (approved):

PR #300, opened 2024-05-02
Competing hazards recovery infection resolution

RJSheppard:
This PR resolves the competing hazards of infection and recovery. Giovanni wrote code to generate two competing hazards objects: CompetingOutcome (in which a function describing what happens to individuals following competing hazard resolution and the rates that that outcome occurs) and CompetingHazard (which links to each of the CompetingOutcome objects that require resolution, and which contains the process function that resolves the competing hazards), found in competing_hazards.R

We first replace create_biting_process and create_progression_process(es) with create_infection_rates_process and create_recovery_rate_process.

Create_infection_rate_process uses the renamed “biting_rate” (which was previously: create_biting_rate_process) which continues as previously, passing through the infection outcome object, and now ends when the rates of infection are generated. Note that we are currently only resolving infections in SA and U compartments, NOT D or Tr (consistent with deterministic model).  Note also that new infections are not rendered until after occurrence of resolution.

Similarly, all recovery/progression processes are now replaced with a single create_recovery_rates_process function, that uses calculate_recovery_rates, based on individual state rates of recovery.

The create_infection_recovery_hazard_process then uses the CompetingHazard object (to which each competing hazard outcome is assigned) to resolve competing hazards and determine which outcome occurs to which individuals.

The infection_outcome uses a new function: infection_process_resolved_hazard, which picks up from where we previously left off in the infection process, rendering incidence and assigning infections severities. Note that this function requires the infection rates to allow the incidence renderer to function.

The recovery_outcome then uses recovery_process_resolved_hazard, streamlining the progression functions into the relevant update functions. This function also passes in the dt variable for antimalarial resistance @tbreweric. (Hopefully it will still work!).

I ran simulations to compare the current default model (with mortality as the last process) with the resolved hazards. We do see something of an increase in incidence (which increases with EIR) and prevalence.
[competing_hazards_mortality_process_position.pdf](https://github.com/mrc-ide/malariasimulation/files/15190828/competing_hazards_mortality_process_position.pdf)
[competing_hazards_resolved_timesteps.pdf](https://github.com/mrc-ide/malariasimulation/files/15190830/competing_hazards_resolved_timesteps.pdf)

Human states also see fewer susceptibles, and is possibly deviating from the equilibrium a little.
[competing_hazards_resolved_states.pdf](https://github.com/mrc-ide/malariasimulation/files/15190885/competing_hazards_resolved_states.pdf)
[competing_hazards_resolved_states_time.pdf](https://github.com/mrc-ide/malariasimulation/files/15190895/competing_hazards_resolved_states_time.pdf)

Acquired immunities seem to be increasing a bit through time - consistent with higher onward infection following higher overall prevalence, particularly at lower EIR (higher plots).
[competing_hazards_resolved_immunity.pdf](https://github.com/mrc-ide/malariasimulation/files/15190880/competing_hazards_resolved_immunity.pdf)

I also checked through the expected infections and recoveries compared with the equilibrium solution, and found reasonable consistency, and resolution of the hazards. Let me know your thoughts!


R/disease_progression.R
```diff
@@ -1,3 +1,79 @@
...
+#' @param recovery_outcome competing hazards object for recovery rates
+#' @noRd
+calculate_recovery_rates <- function(variables, parameters, recovery_outcome, dt_input){
+  
+  # Get correct input for dt depending on antimalarial resistance
+  if(isFALSE(parameters$antimalarial_resistance) & length(dt_input) == 1 & is.numeric(dt_input)){
+    dt_v <- dt_input
+  } else if (isTRUE(parameters$antimalarial_resistance)) {
+    dt_v <- dt_input$get_values(variables$state$get_index_of("Tr"))
+  }
```
giovannic, line 16: We probably want to cover all cases and use boolean instead of bitwise operators.

```suggestion
  if(isFALSE(parameters$antimalarial_resistance)){
    stopifnot(length(dt_input) == 1 && is.numeric(dt_input)) # please move this check to pre-simulation time for efficiency
    dt_v <- dt_input
  } else {
    dt_v <- dt_input$get_values(variables$state$get_index_of("Tr"))
  }
```

R/disease_progression.R
```diff
@@ -1,3 +1,79 @@
...
+
+  recovery_rates <- numeric(length = parameters$human_population)
+  recovery_rates[variables$state$get_index_of("D")$to_vector()] <- 1/parameters$dd
+  recovery_rates[variables$state$get_index_of("A")$to_vector()] <- 1/parameters$da
+  recovery_rates[variables$state$get_index_of("U")$to_vector()] <- 1/parameters$du
+  recovery_rates[variables$state$get_index_of("Tr")$to_vector()] <- 1/dt_v
+  recovery_outcome$set_rates(recovery_rates)
```
giovannic, line 23: It's perhaps more efficient to update variable called `recovery_rates` in the progression processes. Then all of this becomes.

```suggestion
  recovery_outcome$set_rates(variables$recovery_rates$get_values())
```

R/competing_hazards.R
```diff
@@ -0,0 +1,82 @@
...
+      }
+      private$targeted_process <- targeted_process
+      self$rates <- rep(0, size)
+    },
+    set_rates = function(rates){
+      self$rates <- rates
+    },
+    stash_source_humans = function(source_humans){
+      self$source_humans <- source_humans
+    },
```
giovannic, line 23: We don't need this anymore? It's ugly 😆 

```suggestion
```

R/human_infection.R
```diff
@@ -1,23 +1,26 @@
 #' @title Simulate malaria infection in humans
 #' @description
-#' Updates human states and variables to represent asymptomatic/clinical/severe
-#' and treated malaria; and resulting boosts in immunity
+#' This function ends with the assignment of rates of infection to the competing
+#' hazard resolution object.  Boosts immunity given infectious bites.
```
giovannic, line 4:
```suggestion
#' hazard resolution object and boosts immunity given infectious bites.
```

R/human_infection.R
```diff
@@ -29,85 +32,35 @@ simulate_infection <- function(
...
+    variables,
+    bitten_humans,
+    parameters,
+    renderer,
+    timestep,
+    infection_outcome
+) {
   source_humans <- variables$state$get_index_of(
     c('S', 'A', 'U'))$and(bitten_humans)
+  infection_outcome$stash_source_humans(source_humans)
```
giovannic, line 63:
```suggestion
```

R/human_infection.R
```diff
@@ -155,21 +108,107 @@ calculate_infections <- function(
...
+#' and treated malaria; and resulting boosts in immunity
+#' @param timestep current timestep
+#' @param infected_humans bitset of infected humans
+#' @param source_humans bitset of humans with infection potential
+#' @param variables a list of all of the model variables
+#' @param renderer model render object
+#' @param parameters model parameters
+#' @param prob vector of population probabilities of infection
+#' @noRd
+infection_process_resolved_hazard <- function(
```
giovannic, line 142: Just a suggestion, this feels easier to understand

```suggestion
infection_outcome_process <- function(
```

R/human_infection.R
```diff
@@ -468,7 +523,7 @@ severe_immunity <- function(age, acquired_immunity, maternal_immunity, parameter
   parameters$theta0 * (parameters$theta1 + (1 - parameters$theta1) / (
     1 + fv * (
       (acquired_immunity + maternal_immunity) / parameters$iv0) ** parameters$kv
-    )
+  )
```
giovannic, line 526: This whitespace actually helps to match brackets

```suggestion
    )
```

R/model.R
```diff
@@ -119,6 +119,7 @@ run_resumable_simulation <- function(
   events <- create_events(parameters)
   initialise_events(events, variables, parameters)
   renderer <- individual::Render$new(timesteps)
+  populate_incidence_rendering_columns(renderer, parameters)
```
giovannic, line 122: Can we add this to `run_simulation` and `run_metapop_simulation` too?

R/processes.R
```diff
@@ -61,16 +61,38 @@ create_processes <- function(
...
     )
   }
+  
+  # =====================================================
+  # Competing Hazard Outcomes (Infections and Recoveries)
+  # =====================================================
+  
+  infection_outcome <- CompetingOutcome$new(
+    targeted_process = function(timestep, target){
+      infection_process_resolved_hazard(timestep, target, infection_outcome$source_humans,
```
giovannic, line 71:
```suggestion
      infection_outcome_process(timestep, target,
```

R/processes.R
```diff
@@ -61,16 +61,38 @@ create_processes <- function(
...
+                                        variables, renderer, parameters, prob = rate_to_prob(infection_outcome$rates))
+    },
+    size = parameters$human_population
+  )
+  
+  recovery_outcome <- CompetingOutcome$new(
+    targeted_process = function(timestep, target){
+      recovery_process_resolved_hazard(timestep, target, variables, parameters, renderer)
```
giovannic, line 79:
```suggestion
      recovery_outcome_process(timestep, target, variables, parameters, renderer)
```

R/processes.R
```diff
@@ -61,16 +61,38 @@ create_processes <- function(
...
   # ==============================
-  # schedule infections for humans and set last_boosted_*
+  # simulate infections for humans and set last_boosted_*
   # move mosquitoes into incubating state
   # kill mosquitoes caught in vector control
-  processes <- c(
-    processes,
-    create_biting_process(
+  create_infection_rates_process <- function(timestep){
+    biting_process(
```
giovannic, line 95: couldn't we keep the function factory in biting_process.R?

In this file we've been calling function factories from *_process.R files. But now we're making the functions in processes.R. It feels a bit inconsistent. And verbose since we have a lot of processes to create.

R/processes.R
```diff
@@ -121,18 +123,23 @@ create_processes <- function(
...
-      'Tr',
-      'S',
-      dt_input,
-      variables$infectivity,
-      0
+  # ===================
+  # Disease Progression
+  # ===================
+  
+  create_recovery_rates_process <- function(timestep){
```
giovannic, line 130: Again, making the function here when it could be in disease_progression.R

R/human_infection.R
```diff
@@ -155,21 +108,107 @@ calculate_infections <- function(
...
+#' @param source_humans bitset of humans with infection potential
+#' @param variables a list of all of the model variables
+#' @param renderer model render object
+#' @param parameters model parameters
+#' @param prob vector of population probabilities of infection
+#' @noRd
+infection_process_resolved_hazard <- function(
+    timestep,
+    infected_humans,
+    source_humans,
```
giovannic, line 145: This is not used

```suggestion
```

giovannic (changes requested):
This is a massive change. Thank you so much for the effort, it has been really thorough.

Few changes for readability and performance

giovannic:
Also if you could rebase with dev, then we can run benchmarks to get an idea of the performance tradeoff.

pwinskill (commented):
Thanks @RJSheppard. Some additional comments from me.

I was looking at the plots you submitted with the initial PR. Are these still valid now? The increase in prevalence for mid-EIRs was quite marked (if I was interpreting the plots correctly), which I think we should check and be cautious about.

I think, if those changes are that apparent we should probably look at PfPr~EIR plot similar to that in Penny et al. If possible lines for current and after changes would be helpful. 

image

R/render.R
```diff
@@ -169,9 +189,20 @@ create_age_group_renderer <- function(
     renderer
 ) {
   function(timestep) {
-    for (i in seq_along(parameters$age_group_rendering_min_ages)) {
-      lower <- parameters$age_group_rendering_min_ages[[i]]
-      upper <- parameters$age_group_rendering_max_ages[[i]]
+
```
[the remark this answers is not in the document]
giovannic: Yes! I thought I commented this, but that's gone missing. Good catch

R/render.R
```diff
@@ -69,41 +64,66 @@ create_prevelance_renderer <- function(
...
+#' @param renderer object for model outputs
+#' @param source_pop the population which is sampled for infection (bitten SAU for incidence, infecte for clinical/severe)
+#' @param prob probability of infection
+#' @param prefix for model outputs
+#' @param lowers age bounds
+#' @param uppers age bounds
+#' @param timestep current target
+#' 
+#' @noRd
+incidence_probability_renderer <- function(
```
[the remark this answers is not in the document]
RJSheppard: Whenever I look at this it always takes me a few to work out what it even means. I'd be happy to remove it, but agree it would be good to check if anyone uses it.

R/disease_progression.R
RJSheppard: I've made this change.

R/disease_progression.R
RJSheppard: I think this actually tidies up some of the issues I was having integrating with @tbreweric 's dt variable. I'm guessing that if we're storing all the recovery rates as a variable then we can replace the dt variable with a recovery_rates variable and still store all the relevant information. Let me know what you think with what I've done.

R/competing_hazards.R
RJSheppard: Yep - gone!

R/human_infection.R
RJSheppard: Made this change.

R/model.R
RJSheppard: So I think run_simulation calls the run_resumable_simulation anyway, so the incidence populate_incidence_rendering_columns should be called there. I've added a wrapper version that loops round the list of rendering objects to the run_metapop_simulation function which I think is working.

R/render.R
```diff
@@ -169,9 +189,20 @@ create_age_group_renderer <- function(
...
+    
+    unique_age_combinations <- as.data.frame(unique(rbind(cbind(parameters$prevalence_rendering_min_ages, parameters$prevalence_rendering_max_ages),
+          cbind(parameters$incidence_rendering_min_ages, parameters$incidence_rendering_max_ages),
+          cbind(parameters$clinical_incidence_rendering_min_ages, parameters$clinical_incidence_rendering_max_ages),
+          cbind(parameters$severe_incidence_rendering_min_ages, parameters$severe_incidence_rendering_max_ages),
+          cbind(parameters$age_group_rendering_min_ages, parameters$age_group_rendering_max_ages))))
+    
+    ordered_unique_age_combinations <- unique_age_combinations[
+      with(unique_age_combinations, order(V1, V2)),
+    ]
```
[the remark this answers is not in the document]
RJSheppard: Added this now - all seems to be working.

R/render.R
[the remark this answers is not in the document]
RJSheppard: I'm not sure I fully understand this. Is it as simple as moving the not repeating parts outside of the time looping section? The actual age rendering needs to occur at each time step, but the age rendering groups only need to be assigned once. I think I've done this now, but please let me know if it isn't what you meant!

RJSheppard:
@pwinskill I'll rebase and make some plots as requested.

RJSheppard:
[competing_hazards_inc_prev.pdf](https://github.com/user-attachments/files/16052882/competing_hazards_inc_prev.pdf)
[competing_hazards_states.pdf](https://github.com/user-attachments/files/16052883/competing_hazards_states.pdf)

Here are my latest comparison plots (@pwinskill) - as far as I can tell any difference has basically disappeared - possibly by increasing pop size and length of simulation.

RJSheppard:
image
Immunity plots

giovannic (commented):
Thanks! All my issues have been addressed.

Please remove the new dead code as commented.

Just going to wait for a benchmark run before passing the final review.

giovannic (approved):

plietar (approved):
I might try to get those missing 40% back later but it doesn't seem worth blocking merging this anymore.

plietar:
All done from my end. There's not much more to be gained I think.

PR #301, opened 2024-05-03
Older age groups (older than equilibrium age object) are not correctl…

RJSheppard:
I noticed that when individuals are older than the max equilbrium solution age group, the human states and immunity assignments doesn't know how to interpret it. It then assigns states and immunity using the youngest age group, which underestimates infection and acquired immunity, while overestimating maternal immunity in this age group (and explains why we often see an initial peak in maternal immunity). I've tried to correct this. Plots show the average over 10 simulations at the first timestep.

State plots (fewer S in age_correction (older)):
image

Immunity plots (slightly higher in acquired imm, lower in maternal):
image

And Immunities through time (averaged over 10 repetitions, EIR = 50):
image

### DIFF
--- a/R/variables.R
+++ b/R/variables.R
@@ -321,8 +321,6 @@ initial_immunity <- function(
   ) {
   if (!is.null(eq)) {
     age <- age / 365
-    # older ages must be assigned within the equilibrium age bounds
-    age[age>=max(eq[[1]][, 'age'])] <- max(eq[[1]][, 'age'])-0.01
     return(vnapply(
       seq_along(age),
       function(i) {
@@ -340,8 +338,6 @@ initial_state <- function(parameters, age, groups, eq, states) {
   if (!is.null(eq)) {
     eq_states <- c('S', 'D', 'A', 'U', 'T')
     age <- age / 365
-    # older ages must be assigned within the equilibrium age bounds
-    age[age>=max(eq[[1]][, 'age'])] <- max(eq[[1]][, 'age'])-0.01
     return(vcapply(
       seq_along(age),
       function(i) {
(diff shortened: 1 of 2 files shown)


giovannic (approved):

R/variables.R
```diff
@@ -321,6 +321,8 @@ initial_immunity <- function(
   ) {
   if (!is.null(eq)) {
     age <- age / 365
+    # older ages must be assigned within the equilibrium age bounds
+    age[age>=max(eq[[1]][, 'age'])] <- max(eq[[1]][, 'age'])-0.01
```
pwinskill, line 325: Would adding a `max_age` parameter to `calculate_initial_ages()` and overwriting any initial ages > max_age in that function be a cleaner and clearer application?

RJSheppard: That sounds reasonable - Michael's model uses a truncated exponential distribution to draw ages anyway, so it doesn't have the same issue. It probably wouldn't be unreasonable to switch over to that. @giovannic ?

giovannic: hmm, it seems right to sample from a truncated exponential here too. From a quick look at mrc-ide/malariasimulation, it would be consistent.

You could use the maximum age in EQUILIBRIUM_AGES as your upper bound. `rtexp` in utils.R could do the sampling for you fyi.

RJSheppard: Sure thing. So if we sample ages from the truncated exponential, there won't be an issue with individuals being outside the eq solution.

My only concern is the way that the initial age distribution would be different to the age distribution at model equilibrium, although maybe this wouldn't be a problem. MW also limits ages to the max equilibrium solution age, so maintains a truncated exp dist throughout, but this might not be consistent with our model...?

RJSheppard:
Here's another plot showing average immunities (I think the ivm is higher in master because Tom corrected the maternal severe immunity parameter). If we're happy, we can merge? (EIR = 50, 20 reps, 10000 pop size). I'm finding it interesting that ib trends up while ica and id trend more down (in all three branches I plotted), but it's reassuring to see that icm and ivm reach the same equilibria fairly quickly across branches.
image

R/variables.R
pwinskill: It should have very minimal impact as we're changing a very small proportion of the population

pwinskill (approved):

PR #304, opened 2024-05-15
Continuous benchmarking

plietar:


plietar:
See https://github.com/plietar/malariasimulation/pull/9 for an example interaction. The link at the end of the bot's comment points to more detailed plots.

The workflow file needs to be merged into master, since that is what GitHub looks for to react to PR comments, which is why this PR targets that branch.

(Coincidentally I forgot I had this branch lying around that I should probably open a real PR for, since the numbers on it look pretty good...)

plietar:
Unfortunately it seems the master branch doesn't build anymore since the latest release to dqrng, which is why CI is now failing. We need to cherry-pick some of the fixes I've made on the `dev` branch onto master. Will make a separate PR for that.

pwinskill (approved):
Thanks so much @plietar! This looks like a really neat solution.
At some point I will sit down with others in the group and work out a sensible (resonably minimal) set of runs that we might want to benchmark against.

giovannic (approved):
V cool!

giovannic:
Is this good to go? I would really like this for #300

PR #315, opened 2024-07-03
Vivax competing hazards main parameters

RJSheppard:
Added parameters csv file that contains falciparum and vivax specific parameter sets.

These are now selected in the get_parameters() function by specificying parasite == "falciparum" or "vivax". The functions then selects the parasite-specific values and combines then withe the non-parasite specific values.

In order to allow the vivax code to run without full functionality, we have included additional falciparum parameters in the vivax parameter set. These will be removed in the following stages.

Added tests to check parameters are behaving as expected.


data-raw/parasite_parameters.R
```diff
@@ -0,0 +1,9 @@
+## code to prepare parasite parameters
+parasite_csv <- read.csv("data-raw/parasite_parameters.csv")
```
pwinskill, line 2: Is this a good spot to add some details about `parasite-parameters.csv`: where is it from, author, date etc?

tests/testthat/test-vivax.R
```diff
@@ -0,0 +1,26 @@
...
+test_that('Test parasite = vivax sets parameters$parasite to vivax', {
+  vivax_parameters <- get_parameters(parasite = "vivax")
+  expect_identical(vivax_parameters$parasite, "vivax")
+})
+
+test_that('Test difference between falciparum and vivax parameter lists', {
+  falciparum_parameters <- get_parameters(parasite = "falciparum")
+  vivax_parameters <- get_parameters(parasite = "vivax")
+  
+  expect_true(all(names(falciparum_parameters)[!names(falciparum_parameters) %in% names(vivax_parameters)] %in%
```
pwinskill, line 22: This I suspect isn't doing what you expect. 
`names(falciparum_parameters)[!names(falciparum_parameters) %in% names(vivax_parameters)]` is `character(0)`.
I'm not 100% sure what you're testing, so more explicit steps would help. I've re-written some as an example here:

```
test_that('Test difference between falciparum and vivax parameter lists', {
  falciparum_parameters <- get_parameters(parasite = "falciparum")
  vivax_parameters <- get_parameters(parasite = "vivax")
  
  vivax_specific_parameters <- c(
    "du","rvm","rva","rb","b0","b1","ib0","kb","theta0","theta1","kv","fv0",
    "av","gammav","iv0","fd0","ad","gammad","d1","id0","kd","ub","uv","gamma1",
    "pvm","init_ivm","init_ib","init_iva"
  )
  
  in_vivax_not_falciparum <- setdiff(names(vivax_parameters), names(falciparum_parameters))
  expect_identical(
    sort(in_vivax_not_falciparum),
    c("alm50", "apcr50", "b", "ca", "dpcr_max", "dpcr_min", "f", 
      "gammal", "init_hyp", "init_iaa", "init_iam", "klm", "kmax", 
      "kpcr", "philm_max", "philm_min", "ra", "ua")
  )
})
```

tests/testthat/test-vivax.R
```diff
@@ -0,0 +1,26 @@
...
+  parameters_def <- get_parameters()
+  parameters_fal <- get_parameters(parasite = "falciparum")
+  expect_identical(parameters_def, parameters_fal)
+})
+
+test_that('Test vivax model runs', {
+  vivax_parameters <- get_parameters(parasite = "vivax")
+  sim_res <- run_simulation(timesteps = 100, parameters = vivax_parameters)
+  expect_equal(nrow(sim_res), 100)
+})
```
pwinskill, line 11: Is this test required? I expect there are other tests that run the model with vivax parameters that would fail if this one did?

tests/testthat/test-vivax.R
pwinskill: Might also be worth thinking about how this would deal with future `Pf` parameter changes/additions - it might be quite unexpected for the dev to have to change `test-vivax.R` if making a `Pf` parameter name change for example?

pwinskill (changes requested):

data-raw/parasite_parameters.R
RJSheppard: There are some parameter details on the get_parameters function help page, which I think is more accessible to users? I'm happy to add them in here too, if you think that would be good.

tests/testthat/test-vivax.R
RJSheppard: I've addressed the tests as we discussed.

pwinskill (changes requested):
Hi Richard,

Currently I can run this with no warnings:

```
vivax_parameters <- get_parameters(parasite = "vivax") |>
  malariasimulation::set_parameter_draw(draw = 10)
```

Am I right in thinking that this is incorrect and that we currently don't have parameter draws for vivax? If so, can you add a error in `set_parameter_draw()` to check for this?

RJSheppard:
@pwinskill This is a good spot - I think I might have the correct parameter draws data sets somewhere, but need to dig this out and review to make sure it's what I think it should be... (Some of the column headings were mislabeled in the file I'd been sent).

giovannic (approved):
Once Pete's comments have been addressed, this looks good to me!

giovannic:
I wouldn't want to hold back this PR by discussing the best way to benchmark vivax without adding an hour to our benchmark task. Could we move that functionality to a separate PR?

Otherwise, we can merge this when @pwinskill is happy with it.

RJSheppard:
I've taken out that functionality and tried to rebase again. I'm hoping that everything works now and that it's ready for merging, but I'm suspicious that it doesn't seem to be merging into the dev branch at the minute? Lots of files seem to have been changed, but these look like they are consistent with dev anyway, so I'm not sure what they're doing...?

pwinskill (approved):

PR #316, opened 2024-07-03
p.v constant asymptomatic infectivity

RJSheppard:
p.v has a constant asymptomatic infectivity, in contrast to p.f where infectivity is affected by age and immunity. We must therefore change this when asymptomatic infectivity is assigned (when an individual become asymptomatic). This occurs in initial assignments, when someone becomes infected to an asymptomatic status infection and when a clinical diseased individual recovers to asymptomatic. We can now use the existing update_infection() function for vivax in human_infection and in disease progression.

We now remove the gamma1 parameter from the vivax parameter list, used in the falciparum code.


R/human_infection.R
```diff
@@ -464,12 +464,26 @@ schedule_infections <- function(
...
+    } else if (parameters$parasite == "vivax"){
+      # p.v has constant asymptomatic infectivity
+      update_infection(
+        variables$state,
+        'A',
+        variables$infectivity,
+        parameters$ca,
+        variables$recovery_rates,
+        1/parameters$da,
+        to_infect
```
pwinskill, line 484: Should this be `to_infect_asym`?

tests/testthat/test-vivax.R
```diff
@@ -24,3 +24,41 @@ test_that('Test difference between falciparum and vivax parameter lists', {
   expect_true(all(names(vivax_parameters[!names(vivax_parameters) %in% names(falciparum_parameters)]) %in%
                     c("dpcr_max","dpcr_min","kpcr","apcr50","init_iaa","init_iam","ra","ua","b","philm_min","philm_max","klm","alm50","ca","f","gammal","init_hyp","kmax")))
 })
+
+test_that('Test age structure should not change vivax infectivity', {
```
pwinskill, line 28: I think I understand what you're getting at with this test, but could it be done more simply just by setting the `variables$id` value for your pop, then checking what `new_infectivity` has been set? I think that would should highlight that Pf depends on the variable and Pv doesn't?

RJSheppard: This one I find a bit trickier. The falciparum code has a function to calculate asymptomatic_infectivity (where you could set id), but the vivax doesn't have a comparable function because it's constant. The options for outer functions to run that generate infectivity from variables are in create_variables, recovery_outcome_process and schedule_infections, but all these are quite bulky - there isn't really a neat way of checking immunity -> infectivity, as far as I can tell.

giovannic (approved):
LGTM

PR #320, opened 2024-07-03
Allow p.v subpatent infections

RJSheppard:
p.v allows subpatent infections to occur, so we now calculate patent subpatent infections, render these and schedule them.

### DIFF
--- a/R/human_infection.R
+++ b/R/human_infection.R
@@ -154,17 +154,19 @@ infection_outcome_process <- function(
     parameters,
     prob){
   
-  incidence_renderer(
-    variables$birth,
-    renderer,
-    infected_humans,
-    'inc_',
-    parameters$incidence_rendering_min_ages,
-    parameters$incidence_rendering_max_ages,
-    timestep
-  )
-  
   if (infected_humans$size() > 0) {
+    
+    renderer$render('n_infections', infected_humans$size(), timestep)
+    incidence_renderer(
+      variables$birth,
+      renderer,
+      infected_humans,
+      'inc_',
+      parameters$incidence_rendering_min_ages,
+      parameters$incidence_rendering_max_ages,
+      timestep
+    )
+    
     boost_immunity(
       variables$ica,
       infected_humans,
@@ -172,6 +174,7 @@ infection_outcome_process <- function(
       timestep,
       parameters$uc
     )
+
     if(parameters$parasite == "falciparum"){
       boost_immunity(
         variables$id,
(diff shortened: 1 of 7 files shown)


R/human_infection.R
```diff
@@ -229,6 +249,53 @@ infection_outcome_process <- function(
...
+  
+  iaa <- variables$iaa$get_values(infections)
+  iam <- variables$iam$get_values(infections)
+  
+  philm <- anti_parasite_immunity(
+    min = parameters$philm_min, max = parameters$philm_max, a50 = parameters$alm50,
+    k = parameters$klm, iaa = iaa, iam = iam)
+  patent_infections <- bitset_at(infections, bernoulli_multi_p(philm))
+  
+  incidence_renderer(
```
pwinskill, line 277: Might be worth chatting through this in person - I'm not clear what `patent_incidence` is really and how it would be used.

pwinskill (commented):
Let's chat about this one @RJSheppard  - I think we want to be careful not to imply things are not modelling in Pf, by the fact they are explicitly modelled in Pv

R/human_infection.R
```diff
@@ -180,53 +184,136 @@ infection_outcome_process <- function(
...
+        variables,
+        clinical,
+        parameters,
+        timestep,
+        renderer
+      )
+      
+      to_U <- infected_humans$and(lm_detectable$not(F))$and(variables$state$get_index_of(c("S")))
+      to_A <- lm_detectable$and(clinical$not(F))
+      to_D <- clinical$and(treated$not(F))
```
pwinskill, line 258: I wonder if a bitset is getting erroneously overwritten somewhere here?

if I add in this check 
``` 
      ih_s <- infected_humans$size()
      to_U <- infected_humans$and(lm_detectable$not(F))$and(variables$state$get_index_of(c("S")))
      to_A <- lm_detectable$and(clinical$not(F))
      to_D <- clinical$and(treated$not(F))
      if(to_D$size() + to_A$size() + to_U$size() != ih_s){
        browser()
      }
```
It gets triggered when I run the model.
I presume that that sum of people going to these states should = the number infected?

pwinskill: It was similar in the Pf part too.

pwinskill (changes requested):
Thanks Richard.
I'll really like these re-formatting, it is much clearer to me what is going on now. 
I've one comment regarding some suspicious behaviour

R/human_infection.R
RJSheppard: I don't think this is quite right. You're right in that infected_humans, lm_detectable and clinical all get replaced when they get called by the new subset, but this should be ok because they aren't called again following their respective assignments.

The original infected_humans bitset also includes the "to_T" assignments, which get assigned elsewhere (in the calculate_treated function). Is this the cause of the issue?

R/human_infection.R
```diff
@@ -540,9 +603,29 @@ schedule_infections <- function(
         'A',
         variables$infectivity,
         parameters$ca,
-        variables$progression_rates,
+        variables$recovery_rates,
```
giovannic, line 606: Judging from [above](https://github.com/mrc-ide/malariasimulation/pull/320/files#diff-12d8549a471c17bfc3f717bb09830ff339757ca9a0362eb56e41ee47b346a9c1L521), "progression_rates" hasn't been renamed to "recovery_rates" in this version

giovannic (approved):
Great! Thank you!

PR #321, opened 2024-07-03
Add hypnozoites to the P.v model

RJSheppard:
We must model hynozoite batch formation via bite infections, decay of hypnozoite batches through, new infections via relapse and the reseting of batch number to 0 for new births.

We create a hypnozoite variable.

The vivax model considers the number of new bites an individual receives (in contrast to whether an individual has been bitten regardless of number of bites) to calculate the infection rates. This is now captured in bitten_humans, which is now a list that also includes n_bites_each.

The vivax model does not factor the eir by the population level biting heterogeneity (although it does use the heterogeneity to distribute biting, but retains the overall total biting according to the EIR). biting_rate.R

In vivax, we track bites in all individuals, not just SAU as in falciparum. This is because new hypnozoite batches can occur in any human state, regardless of whether this results in a new infection.

The number of bites is considered in calculating probability of a new infection.

To calculate the infection rates for vivax, we take the biting infection rates and add them to the relase rate. We also save the relative rates in the competing hazards object to resolve these competing hazards later.

A new function to resolve between bites and relapse infections is created. For each bite we increase hte number of batches - with capacity for prophylaxis to be added at a future step. We also cap the total number of batches.

We add a hypnozoite batch decay process. Note that this process differs from the immunity decay processes. We work in discrete outputs (you can't have a fractional batch). We also work in the probability that an individual experiences a single loss, rather than loss at the hypnozoite level.

When an individual dies, we reset the hypnozoite batches back to 0.

We now render number of relapses, average hypnozoite batch number and the number of individuals with hypnozoite batches. We may also output these by specified age groups.

Added a test to check relapses occur as expected.

### DIFF
--- a/R/biting_process.R
+++ b/R/biting_process.R
@@ -34,7 +34,7 @@ create_biting_process <- function(
   function(timestep) {
     # Calculate combined EIR
     age <- get_age(variables$birth$get_values(), timestep)
-    bitten_humans <- simulate_bites(
+    bitten <- simulate_bites(
       renderer,
       solvers,
       models,
@@ -52,7 +52,8 @@ create_biting_process <- function(
     simulate_infection(
       variables,
       events,
-      bitten_humans,
+      bitten$bitten_humans,
+      bitten$n_bites_per_person,
       age,
       parameters,
       timestep,
@@ -78,6 +79,7 @@ simulate_bites <- function(
   mixing_index = 1
   ) {
   bitten_humans <- individual::Bitset$new(parameters$human_population)
+  n_bites_per_person <- numeric(0)
   
   human_infectivity <- variables$infectivity$get_values()
   if (parameters$tbv) {
@@ -146,13 +148,24 @@ simulate_bites <- function(
 
     renderer$render(paste0('EIR_', species_name), species_eir, timestep)
     EIR <- EIR + species_eir
-    expected_bites <- species_eir * mean(psi)
+    if(parameters$parasite == "falciparum"){
+      # p.f model factors eir by psi
+      expected_bites <- species_eir * mean(psi)
+    } else if (parameters$parasite == "vivax"){
+      # p.v model standardises biting rate het to eir
+      expected_bites <- species_eir
+    }
+    
     if (expected_bites > 0) {
       n_bites <- rpois(1, expected_bites)
       if (n_bites > 0) {
-        bitten_humans$insert(
-          fast_weighted_sample(n_bites, lambda)
-        )
+        bitten <- fast_weighted_sample(n_bites, lambda)
+        bitten_humans$insert(bitten)
+        renderer$render('n_bitten', bitten_humans$size(), timestep)
+        if(parameters$parasite == "vivax"){
+          # p.v must pass through the number of bites per person
+          n_bites_per_person <- tabulate(bitten, nbins = length(lambda))
+        }
       }
     }
 
@@ -202,9 +215,8 @@ simulate_bites <- function(
       )
     }
   }
-  
-  renderer$render('n_bitten', bitten_humans$size(), timestep)
-  bitten_humans
+
+  list(bitten_humans = bitten_humans, n_bites_per_person = n_bites_per_person)
 }
 
 
--- a/R/competing_hazards.R
+++ b/R/competing_hazards.R
@@ -25,7 +25,7 @@ CompetingOutcome <- R6::R6Class(
       self$args <- list(...)
     },
     execute = function(t, target){
-      do.call(private$targeted_process, list(t, target, self$args))
+      do.call(private$targeted_process, c(t, list(target), self$args))
     },
     reset = function() {
       self$target$clear()
--- a/R/competing_hazards.R
+++ b/R/competing_hazards.R
@@ -16,31 +16,25 @@ CompetingOutcome <- R6::R6Class(
 
       self$target <- individual::Bitset$new(size)
       self$rates <- NULL
-      self$relative_rates <- NULL
+      self$args <- NULL
     },
-    set_rates = function(target, rates){
+    set_rates = function(target, rates, ...){
       stopifnot(target$size() == length(rates))
-
       self$target$copy_from(target)
       self$rates <- rates
-    },
-    set_relative_rates = function(target, relative_rates){
-      stopifnot(target$size() == length(relative_rates))
-      
-      self$target$copy_from(target)
-      self$relative_rates <- relative_rates
+      self$args <- list(...)
     },
     execute = function(t, target){
-      private$targeted_process(t, target)
+      do.call(private$targeted_process, list(t, target, self$args))
     },
     reset = function() {
       self$target$clear()
       self$rates <- NULL
-      self$relative_rates <- NULL
+      self$args <- NULL
     },
     target = NULL,
     rates = NULL,
-    relative_rates = NULL
+    args = NULL
   )
 )
 
--- a/R/competing_hazards.R
+++ b/R/competing_hazards.R
@@ -16,6 +16,7 @@ CompetingOutcome <- R6::R6Class(
 
       self$target <- individual::Bitset$new(size)
       self$rates <- NULL
+      self$relative_rates <- NULL
     },
     set_rates = function(target, rates){
       stopifnot(target$size() == length(rates))
@@ -23,15 +24,23 @@ CompetingOutcome <- R6::R6Class(
       self$target$copy_from(target)
       self$rates <- rates
     },
+    set_relative_rates = function(target, relative_rates){
+      stopifnot(target$size() == length(relative_rates))
+      
+      self$target$copy_from(target)
+      self$relative_rates <- relative_rates
+    },
     execute = function(t, target){
       private$targeted_process(t, target)
     },
     reset = function() {
       self$target$clear()
       self$rates <- NULL
+      self$relative_rates <- NULL
     },
     target = NULL,
-    rates = NULL
+    rates = NULL,
+    relative_rates = NULL
   )
 )
 
--- a/R/human_infection.R
+++ b/R/human_infection.R
@@ -5,6 +5,7 @@
 #' @param variables a list of all of the model variables
 #' @param events a list of all of the model events
 #' @param bitten_humans a bitset of bitten humans
+#' @param n_bites_per_person vector of number of bites each person receives (p.v only)
 #' @param age of each human (timesteps)
 #' @param parameters of the model
 #' @param timestep current timestep
@@ -15,6 +16,7 @@ simulate_infection <- function(
     variables,
     events,
     bitten_humans,
+    n_bites_per_person,
     age,
     parameters,
     timestep,
@@ -37,6 +39,7 @@ simulate_infection <- function(
   calculate_infections(
     variables,
     bitten_humans,
+    n_bites_per_person,
     parameters,
     renderer,
     timestep,
@@ -48,6 +51,7 @@ simulate_infection <- function(
 #' @description Infection rates are stored in the infection outcome competing hazards object
 #' @param variables a list of all of the model variables
 #' @param bitten_humans bitset of bitten humans
+#' @param n_bites_per_person vector of number of bites each person receives (p.v only)
 #' @param parameters model parameters
 #' @param renderer model render object
 #' @param timestep current timestep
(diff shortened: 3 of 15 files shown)


R/human_infection.R
```diff
@@ -208,57 +259,177 @@ infection_outcome_process <- function(
...
+  
+  ## all bitten humans with an infectious bite (incorporating prophylaxis) get a new batch of hypnozoites
+  if(bite_infections$size()>0){
+    new_hypnozoite_batch_formed <- bite_infections # bitset_at(bite_infections, bernoulli_multi_p(1-ls_prophylaxis))
+    
+    # make sure batches are capped
+    current_batches <- variables$hypnozoites$get_values(new_hypnozoite_batch_formed)
+    new_batch_number <- ifelse(current_batches == parameters$kmax,
+                               current_batches,
+                               current_batches + 1)
```
pwinskill, line 354: I think the following is more concise and a little more efficient?    

```suggestion
    current_batches <- variables$hypnozoites$get_values(new_hypnozoite_batch_formed)
new_batch_number <- pmin(current_batches + 1, parameters$kmax)
```

R/human_infection.R
```diff
@@ -208,57 +259,177 @@ infection_outcome_process <- function(
...
+
```
pwinskill, line 415: See comment above

R/human_infection.R
```diff
@@ -272,28 +443,28 @@ calculate_patent_infections <- function(
...
-    parameters$patent_incidence_rendering_min_ages,
-    parameters$patent_incidence_rendering_max_ages,
+    'inc_lm_det_',
+    parameters$lm_det_incidence_rendering_min_ages,
+    parameters$lm_det_incidence_rendering_max_ages,
     timestep
   )
-  patent_infections
+  lm_det_infections
 }
```
pwinskill, line 468: I think I may have posted a similar comment before, but I don't understand what this output would be used for? With a diagnostic like Light microscopy, the inutitive measure is one of prevalence.

R/parameters.R
```diff
@@ -467,8 +467,8 @@ get_parameters <- function(overrides = list(), parasite = "falciparum") {
       age_group_rendering_max_ages = numeric(0),
       incidence_rendering_min_ages = numeric(0),
       incidence_rendering_max_ages = numeric(0),
-      patent_incidence_rendering_min_ages = numeric(0),
-      patent_incidence_rendering_max_ages = numeric(0),
+      lm_det_incidence_rendering_min_ages = numeric(0),
+      lm_det_incidence_rendering_max_ages = numeric(0),
```
pwinskill, line 471: See comment above

R/processes.R
```diff
@@ -344,3 +362,19 @@ create_lagged_eir <- function(variables, solvers, parameters) {
...
+#' @description
+#' calulates the number of individuals in whom a batch decay occurs
+#'
+#' @param variable the hypnozoite variable to update
+#' @param rate the hypnozoite decay rate
+#' @noRd
+create_hypnozoite_batch_decay_process <- function(hypnozoites, gammal){
+  function(timestep){
+    to_decay <- bernoulli_multi_p(p = rate_to_prob(hypnozoites$get_values() * gammal))
+    hypnozoites$queue_update(
```
pwinskill, line 375: Worth wrapping this in a if statement to not run the update if to_decay is empty?

pwinskill (changes requested):
Hi Richard,
A few Qs to followup on for this one, alongside some suggestions 🌟

R/human_infection.R
```diff
@@ -114,8 +118,26 @@ calculate_infections <- function(
...
+  
+  infection_rates <- rep(0, length = parameters$human_population)
+  if(parameters$parasite == "falciparum"){
+    prob <- b * (1 - prophylaxis) * (1 - vaccine_efficacy)
+    infection_rates[source_vector] <- prob_to_rate(prob)
+    
+  } else if (parameters$parasite == "vivax"){
+    ## calculated rate of infection for all bitten or with hypnozoites
+    relapse_rates <- variables$hypnozoites$get_values() * parameters$f
+    infection_rates[bitten_vector] <- infection_rates[bitten_vector] + prob_to_rate(b)
```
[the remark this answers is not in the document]
RJSheppard: This is an error that I correct in the next PR, but forgot to go back and correct here (I forgot to add the relapses with the bite rates). I'll push a correction.

R/human_infection.R
RJSheppard: Nice - thanks for this one!

R/human_infection.R
RJSheppard: I think this is duplicated code (likely a mistake from rebasing). I've removed this one.

R/human_infection.R
RJSheppard: I'm happy to take this out if you don't think it's useful. I was just trying to give flexible options to the user without thinking much about how it would be used.

R/biting_process.R
```diff
@@ -144,13 +145,23 @@ simulate_bites <- function(
...
       n_bites <- rpois(1, expected_bites)
       if (n_bites > 0) {
-        bitten_humans$insert(
-          fast_weighted_sample(n_bites, lambda)
-        )
+        bitten <- fast_weighted_sample(n_bites, lambda)
+        bitten_humans$insert(bitten)
+        if(parameters$parasite == "vivax"){
+          # p.v must pass through the number of bites each individual receives
+          n_bites_each <- table(bitten)
```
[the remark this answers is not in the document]
RJSheppard: So I was getting around that by passing through the full table - with column names indicating the individuals that had been bitten. But I think using tabulate can make lots of things a bit simpler. I've had another go with this. Let me know what you think. I've tried to use bitsets where I can.

R/human_infection.R
```diff
@@ -61,20 +61,25 @@ calculate_infections <- function(
     infection_outcome
 ) {
   
-  if(bitten_humans$bitten_humans$size() == 0){return(bitten_humans$bitten_humans)}
-  
   if(parameters$parasite == "falciparum"){
-    source_humans <- variables$state$get_index_of(c('S','A','U'))$and(bitten_humans$bitten_humans)
+    
+    if(bitten_humans$size() == 0){return(bitten_humans)}
```
pwinskill, line 66: This line isn't clear to me, can't we just wrap this section in a more classic if statement like:
```
if(bitten_humans$size() > 0){
Do this stuff
}
```

R/human_infection.R
```diff
@@ -61,20 +61,25 @@ calculate_infections <- function(
...
     b <- blood_immunity(variables$ib$get_values(source_humans), parameters)
     
   } else if (parameters$parasite == "vivax"){
     ## source_humans must include individuals with hypnozoites which may be impacted by prophylaxis/vaccination
-    source_humans <- bitten_humans$bitten_humans$copy()$or(variables$hypnozoites$get_index_of(0)$not(TRUE))
-    bitten_vector <- bitten_humans$bitten_humans$to_vector()
+    bitten_humans_index <- bitten_humans$get_index_of(!0)
+    source_humans <- variables$hypnozoites$get_index_of(0)$not(T)$or(bitten_humans_index)
+
+    if(source_humans$size() == 0){return(source_humans)}
```
pwinskill, line 78: Similar to above

R/biting_process.R
```diff
@@ -157,10 +156,15 @@ simulate_bites <- function(
...
+        if(parameters$parasite == "falciparum"){
+          bitten_humans$insert(bitten)
+          renderer$render('n_bitten', bitten_humans$size(), timestep)
+        }
         if(parameters$parasite == "vivax"){
-          # p.v must pass through the number of bites each individual receives
-          n_bites_each <- table(bitten)
+          # p.v must pass through the number of bites per person
+          bites_per_person <- tabulate(bitten, nbins = length(lambda))
+          bitten_humans <- individual::IntegerVariable$new(initial_values = bites_per_person)
```
pwinskill, line 166: I see how this seems tidier, but I have a few concerns.
1. We now have a single variable `bitten_humans` that means different things for Pf and Pv and ...
2. that leads to a mismatch in the `calculate_infections()` function, which takes the argument `bitten_humans()` currently defined as a bitset of bitten humans.

Would it be clearer to:
1. Create the bitset of `bitten_humans` based on `n_bites` and `lambda` for both Pf and Pf and pass this through to `calculate_infections()` as an argument.
2. Additionally create a `IntegerVariable` for vivax called and storing `n_bites`. This can be accessed in `calculate_infections()` directly in that function as you're already passing `variables` as an argument.

pwinskill (changes requested):
...a few more changes suggested

R/biting_process.R
RJSheppard: I've given this a go (and rebased) and I think it's ready for review.

R/human_infection.R
```diff
@@ -145,14 +181,16 @@ calculate_infections <- function(
 #' @param renderer model render object
 #' @param parameters model parameters
 #' @param prob vector of population probabilities of infection
+#' @param relative_rates relative rates of hypnozoite relapse relative to total infection rate
 #' @noRd
 infection_outcome_process <- function(
```
giovannic, line 186: Again, this function is quite long. It would be cleaner as `falciparum_infection_outcome` and `vivax_infection_outcome`.

R/processes.R
```diff
@@ -96,7 +98,8 @@ create_processes <- function(
     targeted_process = function(timestep, target){
       infection_outcome_process(timestep, target, 
                                 variables, renderer, parameters, 
-                                prob = rate_to_prob(infection_outcome$rates))
+                                prob = rate_to_prob(infection_outcome$rates),
+                                relative_rates = infection_outcome$relative_rates)
```
giovannic, line 102: I've only just realised how ugly and genius this is...

I don't yet know how I feel about it

R/competing_hazards.R
```diff
@@ -16,22 +16,31 @@ CompetingOutcome <- R6::R6Class(
...
       self$target <- individual::Bitset$new(size)
       self$rates <- NULL
+      self$relative_rates <- NULL
     },
     set_rates = function(target, rates){
       stopifnot(target$size() == length(rates))
 
       self$target$copy_from(target)
       self$rates <- rates
     },
```
giovannic, line 26: I think a more readable, generalisable - and perhaps less error prone - method would be to explicity accept extra arguments for your outcome when you set rates.

```suggestion
    set_rates = function(target, rates, ...){
      stopifnot(target$size() == length(rates))

      self$target$copy_from(target)
      self$rates <- rates
      self$args <- list(...)
    },
```

R/competing_hazards.R
```diff
@@ -16,22 +16,31 @@ CompetingOutcome <- R6::R6Class(
 
       self$target <- individual::Bitset$new(size)
       self$rates <- NULL
+      self$relative_rates <- NULL
```
giovannic, line 19:
```suggestion
      self$args <- NULL
```

R/competing_hazards.R
```diff
@@ -16,22 +16,31 @@ CompetingOutcome <- R6::R6Class(
...
+      self$target$copy_from(target)
+      self$relative_rates <- relative_rates
+    },
     execute = function(t, target){
       private$targeted_process(t, target)
     },
     reset = function() {
       self$target$clear()
       self$rates <- NULL
+      self$relative_rates <- NULL
```
giovannic, line 39:
```suggestion
      self$args <- NULL
```

R/competing_hazards.R
```diff
@@ -16,22 +16,31 @@ CompetingOutcome <- R6::R6Class(
...
 
       self$target$copy_from(target)
       self$rates <- rates
     },
+    set_relative_rates = function(target, relative_rates){
+      stopifnot(target$size() == length(relative_rates))
+      
+      self$target$copy_from(target)
+      self$relative_rates <- relative_rates
+    },
```
giovannic, line 32:
```suggestion
```

R/competing_hazards.R
```diff
@@ -16,22 +16,31 @@ CompetingOutcome <- R6::R6Class(
...
     },
     reset = function() {
       self$target$clear()
       self$rates <- NULL
+      self$relative_rates <- NULL
     },
     target = NULL,
-    rates = NULL
+    rates = NULL,
+    relative_rates = NULL
```
giovannic, line 43:
```suggestion
    args = NULL
```

R/competing_hazards.R
```diff
@@ -16,22 +16,31 @@ CompetingOutcome <- R6::R6Class(
...
       self$rates <- rates
     },
+    set_relative_rates = function(target, relative_rates){
+      stopifnot(target$size() == length(relative_rates))
+      
+      self$target$copy_from(target)
+      self$relative_rates <- relative_rates
+    },
     execute = function(t, target){
       private$targeted_process(t, target)
```
giovannic, line 34:
```suggestion
      do.call(private$targeted_process, c(c(t, target), self$args)
```

R/processes.R
```diff
@@ -96,7 +98,8 @@ create_processes <- function(
     targeted_process = function(timestep, target){
```
giovannic, line 98:
```suggestion
    targeted_process = function(timestep, target, prob, relative_rates){
```

R/processes.R
giovannic: I think the above edit to CompetingOutcome is a better way to deal with optional arguments for the processes. That way it generalises to different competing hazard scenarios and you don't have to continue adding fields to the R6 object for more arguments.

R/human_infection.R
```diff
@@ -48,91 +51,124 @@ simulate_infection <- function(
...
+      "inc_",
+      parameters$incidence_min_ages,
+      parameters$incidence_max_ages,
+      timestep
     )
+    
+    ## capture infection rates to resolve in competing hazards
+    infection_outcome$set_rates(
+      source_humans, 
+      infection_rates)
```
giovannic, line 170: The edit to CompetingOutcome would allow you to add prob and relative_rates here...

giovannic (changes requested):
Functionally looks fine to me. I have suggested some cleaner ways to organise the code and the competing hazard stuff.

R/human_infection.R
```diff
@@ -15,66 +16,214 @@ simulate_infection <- function(
...
+    relative_rates <- NULL
+    if(hypnozoites_humans$size()>0){
+      relapse_rates <- variables$hypnozoites$get_values(hypnozoites_humans) * parameters$f
+      hyp_source_index <- bitset_index(source_humans, hypnozoites_humans)
+      infection_rates[hyp_source_index] <- infection_rates[hyp_source_index] + relapse_rates
+      # Get and store relative rates for bite/relapse competing hazards resolution
+      relative_rates <- relapse_rates/infection_rates[hyp_source_index]
+      # infection_outcome$set_relative_rates(
+      # hypnozoites_humans, 
+      # relative_rates)
```
giovannic, line 186: dead code

```suggestion
```

R/human_infection.R
```diff
@@ -685,7 +944,7 @@ severe_immunity <- function(age, acquired_immunity, maternal_immunity, parameter
   parameters$theta0 * (parameters$theta1 + (1 - parameters$theta1) / (
     1 + fv * (
       (acquired_immunity + maternal_immunity) / parameters$iv0) ** parameters$kv
-    )
+  )
```
giovannic, line 947:
```suggestion
    )
```

R/competing_hazards.R
```diff
@@ -16,22 +16,25 @@ CompetingOutcome <- R6::R6Class(
...
+    set_rates = function(target, rates, ...){
       stopifnot(target$size() == length(rates))
-
       self$target$copy_from(target)
       self$rates <- rates
+      self$args <- list(...)
     },
     execute = function(t, target){
-      private$targeted_process(t, target)
+      do.call(private$targeted_process, list(t, target, self$args))
```
giovannic, line 28: Oh, I think it's most intuitive to call `targeted_process` with the extra arguments in the same way they were passed to `$set_rates`. i.e.

`outcome$set_rates(rates, a, b)`

should call the process with:

`targeted_process(target, a, b)` <- this version preserves the structure of my extra arguments

not

`targeted_process(target, list(a, b))` <- this version turns my extra arguments into a list.

```suggestion
      do.call(private$targeted_process, c((t, target), self$args))
```

giovannic (changes requested):
Very nearly there for me!

giovannic (approved):
Thanks!

R/human_infection.R
```diff
@@ -270,31 +466,91 @@ infection_outcome_process <- function(
...
+#' 
+#' @param variables a list of all of the model variables
+#' @param infected_humans bitset of infected humans
+#' @param relative_rates relative rate of relapse infection
+#' @param variables model variables
+#' @param parameters model parameters
+#' @param renderer model renderer
+#' @param timestep current timestep
+#' @noRd
+relapse_bite_infection_hazard_resolution <- function(
```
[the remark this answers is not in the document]
RJSheppard: The main CH resolution is between infections and disease progression. This part of the code resolves between relapses and bite infections and doesn't use the competing hazards object. It's relatively simple and discretely resolved, so doesn't need the object.

PR #322, opened 2024-07-03
Add equilibrium solution to P.v model 

RJSheppard:
Added in functionality to start at equilibrium, based on Michael White's equilibrium solution. This is accessed in initial equilibrium calculation in compatability and in variable creation. The equilibrium solution is found in malariaEquilibriumVivax.

Note that we draw individuals from a joint probability distribution over age and het group to generate human state and number of hypnozoite batches.

tests included to check correct equilibrium is generated and maintained.

I have also added the An. koliensis parameters as found in Michael White's repo in order to compare equilibrium solutions.

A few errors were identified. Total infection rates were not being calculated correctly in human_infection.R. Lagged infectivity needs to be saved before the current infectivity is extracted during the biting_process.R. Also, when immunity is boosted through infection, this should incorporate the decrease in immunity from decay (there may be better ways of achieving this).


pwinskill (approved):

tests/testthat/test-vivax-compartmental.R
```diff
@@ -0,0 +1,161 @@
...
+    parameters$total_M
+  )
+  
+  for (t in seq(timesteps)) {
+    expected <- rbind(expected, c(t, equilibrium))
+  }
+  expect_equal(counts, expected, tolerance=1e-3)
+})
+
+test_that('vivax ODE stays at equilibrium with low total_M', {
```
giovannic, line 75: This is perhaps redundant? Is there any change in the vector model for vivax?

tests/testthat/test-vivax-compartmental.R
```diff
@@ -0,0 +1,161 @@
...
+  
+  for (t in seq(timesteps)) {
+    expected <- rbind(expected, c(t, equilibrium))
+  }
+  
+  expect_equal(counts, expected, tolerance=1e-4)
+})
+
+
+test_that('vivax changing total_M stabilises', {
```
giovannic, line 116: Same as above

tests/testthat/test-vivax-eq.R
```diff
@@ -0,0 +1,86 @@
+test_that('Initial states are consistent with equilibrium', {
+  skip_on_ci()
```
giovannic, line 2: We could run this on the CI? It seems valuable and quick to run. If we're not doing this for falciparum, maybe we should be.

tests/testthat/test-vivax-eq.R
```diff
@@ -0,0 +1,86 @@
...
+               vars$state$get_size_of("U"),
+               vars$state$get_size_of("Tr")))
+    })/population
+  
+  expect_equal(object = c(expected_states), expected =  c(actual_states), tolerance = 1E-2)
+  
+})
+
+test_that('Initial immunities are consistent with equilibrium', {
+  skip_on_ci()
```
giovannic, line 38: Same as above

tests/testthat/test-vivax-eq.R
```diff
@@ -0,0 +1,86 @@
...
+               mean(vars$iaa$get_values()),
+               mean(vars$iam$get_values()),
+               mean(vars$hypnozoites$get_values())))})
+  
+  expect_equal(object = c(expected_averages), expected =  c(actual_averages), tolerance = 1E-1)
+  
+})
+
+test_that('vivax equilibrium works with multiple species', {
+  skip_on_ci()
```
giovannic, line 79: Redundant?

giovannic (changes requested):
Very nice! Just making sure the right tests run on CI, comments in line.

tests/testthat/test-vivax-compartmental.R
RJSheppard: I think there was a while when I wasn't sure, but I think you're right: I can take them out.

giovannic (approved):

PR #330, opened 2024-07-30
Vivax competing hazards main drugs

RJSheppard:
Adding drug vivax parameters, including liver stage prophylaxis.


R/human_infection.R
```diff
@@ -729,6 +749,15 @@ calculate_treated <- function(
     )
   }
   
+  # Update liver stage drug effects
+  if(length(parameters$drug_hynozoite_efficacy) > 0){
```
pwinskill, line 753: Are both `if`s required here?
If `successfully_treated$successfully_treated_hypnozoites$size() > 0` wouldn't `length(parameters$drug_hynozoite_efficacy) > 0` by definition?

pwinskill (commented):
Thanks @RJSheppard. 

A few Qs:

- I'm still a bit unclear as to why we have, for exampe both `SP_AQ_params_falciparum` and `SP_AQ_params`. When is the second used?
- I see we can use vivax drugs for MDA. What have we done with SMC and PMC? To my knowledge those aren't avaiable for vivax-specific implementation. Is it worth putting in a check to error if users try to set up either of those interventions with vviax drugs? I can't remeber but perhaps we already have a check to stop those interventions completely if `parasite = vivax`?
- Worth sneeking in a fix for https://github.com/mrc-ide/malariasimulation/issues/339 here?

R/drug_parameters.R
```diff
@@ -1,44 +1,95 @@
...
+#' @details Default parameters, from L to R, are: drug_efficacy: 1, drug_rel_c: 0.5, drug_prophylaxis_shape: 5, drug_prophylaxis_scale: 20,
+#' drug_hypnozoite_efficacy: 0.713, drug_hypnozoite_prophylaxis_shape: 5.5, drug_hypnozoite_prophylaxis_scale: 30
+CQ_TQ_params_vivax <- c(1, 0.5, 5, 20, 0.713, 5, 30)
+
 #' @title Parameterise drugs to use in the model
 #'
 #' @param parameters the model parameters
 #' @param drugs a list of drug parameters, can be set using presets
 #' @export
 set_drugs <- function(parameters, drugs) {
```
giovannic, line 73: Is there any validation we could do here to validate if parameters$parasite == vivax that the correct drugs are used? Or do all drugs make sense for all parasites?

R/human_infection.R
```diff
@@ -519,13 +519,33 @@ relapse_bite_infection_hazard_resolution <- function(
   ## all bitten humans with an infectious bite (incorporating prophylaxis) get a new batch of hypnozoites
   if(bite_infections$size()>0){
     
+    ## drug prophylaxis may limit formation of new hypnozoite batches
+    ls_prophylaxis <- rep(0, bite_infections$size())
+    if(length(parameters$drug_hypnozoite_efficacy) > 0){
+      
+      ls_drug <- variables$ls_drug$get_values(bite_infections)
+      ls_medicated <- (ls_drug > 0)
+      ls_medicated[ls_drug > 0] <- !is.na(parameters$drug_hypnozoite_efficacy[ls_drug])
```
giovannic, line 528: I would expect this to fail? Since `length(ls_medicated[ls_drug > 0])` is less than or equal to `length(ls_drug)`. If some people are not medicated then there'll be a weird assignment here. Is this tested for?

```suggestion
      ls_medicated <- ls_medicated & !is.na(parameters$drug_hypnozoite_efficacy[ls_drug])
```

R/drug_parameters.R
```diff
@@ -2,43 +2,83 @@
...
 
 #' @title Preset parameters for the AL drug
 #' @description From SI of Commun. 5:5606 doi: 10.1038/ncomms6606 (2014)
 #' @details Use a vector of preset parameters for the AL drug (artemether-lumefantrine)
 #' @details Default parameters, from L to R, are: drug_efficacy: 0.95, drug_rel_c: 0.05094, drug_prophylaxis_shape: 11.3, drug_prophylaxis_scale: 10.6
+#' @export
+AL_params_falciparum <- c(.95, 0.05094, 11.3, 10.6)
+
 #' @export
 AL_params <- c(.95, 0.05094, 11.3, 10.6)
```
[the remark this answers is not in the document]
giovannic: I would lean towards keeping AL_params and removing AL_params_falciparum. Bumping the version and getting everyone to change their code is not convenient

giovannic (changes requested):

R/human_infection.R
RJSheppard: I see what you're saying. An issue is that some of ls_drug will be 0 (if no drugs have been assigned to the individual), which means that `!is.na(parameters$drug_hypnozoite_efficacy[ls_drug])` may not have the same length as `ls_medicated` (as far as I can work it out). I've taken a look to make sure it's doing what it's supposed to do, with a few tests to check it through.

R/drug_parameters.R
RJSheppard: I spoke with Pete about this a while ago, and we came to a place of hoping that people would use their good sense to choose which parameter sets to apply for each species. While drug parameters are parasite-specific, there might be circumstances where someone would want to see what happens if the same drug is applied to both species. I've added a few warnings to flag where people might do something strange while maintaining flexibility. I could change some of the warnings to errors (e.g., the length of the parameter inputs) if we think that's important.

R/human_infection.R
RJSheppard: The way I've coded it means that `successfully_treated$successfully_treated_hypnozoites` is only generated when `length(parameters$drug_hynozoite_efficacy) > 0`. So it doesn't exist for falciparum, or for vivax when non hypnozoite drugs are being used (e.g., just CQ).

RJSheppard:
I've removed the `_falciparum` drugs names for backwards compatibility.

vivax drugs can currently be used in SMC/PMC via the update_mass_drug_admin function that is primarily called during MDA (see mda_processes.R). It may not be realistic for this to be used in those settings, but for consistency/flexibility I've left it in. I can imagine circumstances where it might be interesting to look at TQ prophylaxis as SMC/PMC, given it's longer half-life.

I can take a look at the drug vector length issue, but the example code is only causing a problem because the set_drugs function hasn't been used. The parameters are changed manually in the example, which isn't advised. I've added some warnings in set_drugs for when the vector length doesn't match up (and could make these errors if we want to fully restrict people from adding incorrect drug vector lengths).

R/human_infection.R
```diff
@@ -519,17 +519,61 @@ relapse_bite_infection_hazard_resolution <- function(
...
+    timestep
+){
+  
+  ## drug prophylaxis may limit formation of new hypnozoite batches
+  ls_prophylaxis <- rep(0, bite_infections$size())
+  if(length(parameters$drug_hypnozoite_efficacy) > 0){
+
+    ls_drug <- variables$ls_drug$get_values(bite_infections)
+    ls_medicated <- ls_drug > 0
+    ls_medicated[ls_medicated] <- !is.na(parameters$drug_hypnozoite_efficacy[ls_drug])
```
giovannic, line 562: This is quite confusing. And possibly redundant? Don't we expect all liver stage drugs have a hypnozoite efficacy?

giovannic (approved):
Looks good thanks!

I've commented in line on a potentially redundant line, let me know if I'm mistaken.

R/human_infection.R
RJSheppard: If I understand the question, the way I've coded it is that the parameter sets are parasite specific, rather than being blood or liver stage specific. That means that vivax drug parameter sets can be either only blood stage (where they would have 4 parameters), or would have both blood and liver stage parameters (which total 7 parameters), and not all "vivax drugs" have liver stage action. Does that answer the question? (Happy to change things if you have a better way of doing things!)

pwinskill: Would it be more consistent to have liver-stage parameters for each vivax drug, and just set them to 0 if the drug is assumed to have no liver-stage impact?

R/human_infection.R
pwinskill: This might be avoidable if all vivax drugs have that parameter (see other comment)?

pwinskill (commented):
Thanks @RJSheppard  - just a few minor followups from me

PR #341, opened 2024-12-10
PMC age subsetting has a mistake in it. timesteps and time_index in s…

RJSheppard:
I was trying to implement PMC and found that there's a bug that prevents the correct age implementation.


giovannic (approved):
Thank you!

pwinskill (approved):
Well spotted

pwinskill:
As this is a bug, should it be expedited into the main branch somehow?

PR #346, opened 2025-02-05
Update VectorControl_Bednets.Rmd

JDChallenger:
The parameter values used for rn and dn0 were invalid (I think): we require s + d + r = 1, here we had d + r > 1. I've changed the values to represent pyrethroid-only ITNs, at 40% pyrethroid resistance

If I'm right about this, I'm surprised that this scenario runs without causing severe downstream issues (s<0 implies negative mosquito survival?). Could an error message be added somewhere, to prevent this parameter combination from being chosen?


giovannic (approved):
Thank you! I've made an issue to do some validation as well #347

Issue #163, opened 2022-02-15
Add in MAP logistic net-retention function


pwinskill:
Malaria atlas project have a logistic decay in net retention over time (in comparison the the exponential decay currently in malariasimulation). It would be helpfult to have this as a option to align with MAPs country specific fit to data for net usage.

pwinskill:
Example of sampling individual rentention time from the sigmoidal (needs further checking):

```
rnet_retention_time <- function(n, half_life, k = 20) {
  # Safety check
  if (half_life <= 0) stop("half_life must be > 0")
  
  # Time at which all nets fail:
  l <- half_life / sqrt(1 - k / (k - log(0.5)))
  
  # Generate n uniform(0,1) random variables
  U <- runif(n)
  
  # Apply inverse of S(t):
  #  T = l * sqrt( [ -log(U)/k ] / [ 1 - ( -log(U)/k ) ] )
  #  Or more neatly: T = l * sqrt( a / (1 + a ) ),  a = -log(U)/k
  a <- -log(U)/k
  T <- l * sqrt(a / (1 + a))
  
  # Numerically, T will be in [0, l]. If we want to be certain we never
  # exceed l by floating rounding, you can do:
  T <- pmin(T, l)
  
  return(T)
}

set.seed(123)

# Suppose half_life = 365 days
# Generate 10,000 draws
times <- rnet_retention_time(n = 10000, half_life = 365, k = 20)

# Compare to MAP function:
x <- rep(NA, 1000)
for(i in 1:1000){
  x[i] <- (10000 - sum(times <= i)) / 10000
}
plot(x)
lines(netz::net_loss_map(1:1000, half_life = 365, k = 20), col = "red")
```

PR #350, opened 2025-02-14
Implement logistic retention times for nets:

giovannic:
* tests
 * parameterisation
 * updated net throw away schedule


giovannic:
Addresses #163

PR #352, opened 2025-02-19
Epi-outputs error message fix 

RJSheppard:
Fixed bug in incorrect parasite output attribution in epi-outputs function. Added test to check errors work.


pwinskill (approved):
Thanks Richard 🙏

PR #354, opened 2025-03-19
Correcting some issues with vivax liver stage prophylaxis.

RJSheppard:
1. ls_drug and ls_drug_time are now added to the variable list.
2. liver stage drug was not differentiated from blood stage drug, so there was some confusion in subsetting and outputs in calculate_successful_treatments function.

### DIFF
--- a/R/variables.R
+++ b/R/variables.R
@@ -35,8 +35,8 @@
 #' * infectivity - The onward infectiousness to mosquitos
 #' * drug - The last prescribed drug
 #' * drug_time - The timestep of the last drug
-#' * ls_drug - The last prescribed drug
-#' * ls_drug_time - The timestep of the last drug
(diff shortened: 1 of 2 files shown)


R/variables.R
```diff
@@ -338,7 +338,9 @@ create_variables <- function(parameters) {
                    last_boosted_iaa = last_boosted_iaa,
                    iaa = iaa,
                    iam = iam,
-                   hypnozoites = hypnozoites
+                   hypnozoites = hypnozoites,
+                   ls_drug = ls_drug,
```
pwinskill, line 342: Just a note that in the description (lines 38 and 39) it would be worth stating that these a "liver-stage"

pwinskill (approved):
Looks good Richard, thanks for responding to this

PR #355, opened 2025-03-20
Re-generate the R-CMD-check and pkgdown workflows.

plietar:
GitHub has deprecated some of the actions that we were using (namely actions/cache), which broke the builds. Rather than just updating the version pins I've decided to re-generate the files from scratch with `usethis`:

```r
usethis::use_github_action("check-standard")
usethis::use_github_action("pkgdown")
```

As far as I can tell we did not have any bespoke in these actions, and all of the changes come from changes to the `r-lib/actions` repo.


plietar:
@richfitz I don't really have anyone better to review this stuff while Giovanni is away

plietar:
/benchmark

(Obviously there's no useful benchmarking to do here, I just want to see if this still works)

richfitz (approved):

PR #364, opened 2025-06-20
Update set_equilibrium() to overwrite rather than duplicate equilibrium parameters in the parameter list

tbreweric:
Currently, if called twice on the same parameter set the set_equilibrium() will append to separate sets of equilibrium parameters (eq_params, init_EIR, init_foim - init_foim will actually have 3 entries as it already exists in the parameter list before set_equilibrium() is called). 

Although ideally a user wouldn't need to call set_equilibrium() more than once, in practice is can happen. I have proposed a simple change to set_equilibrium() which overwrites the equilibrium parameters (init_foim, init_EIR, and eq_params) if the function is called more than once and added some simple unit tests.

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -353,15 +353,20 @@ run_metapop_simulation <- function(
 #'
 #' @param timesteps the number of timesteps to run the simulation for
 #' @param repetitions n times to run the simulation
-#' @param overrides a named list of parameters to use instead of defaults
+#' @param parameters a named list of parameters to use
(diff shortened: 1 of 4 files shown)


RJSheppard (changes requested):

tbreweric:
Following advice of @RJSheppard I have moved the lines assigning `init_foim` and `init_EIR` to `parameters` outside of the species loops to remove duplicated lines & simplify!

PR #365, opened 2025-06-20
Update set_parameter_draw() to append draw to the parameter list

tbreweric:
It would be useful to be able to easily see whether:

a) a parameter list has been run through `set_parameter_draw()`
b) if so which parameter draw has been used

I have drafted a version of `set_parameter_draw()` which appends the draw provided to the parameter list as `$parameter_draw` and added some basic unit tests for both falciparum and vivax.

### DIFF
--- a/R/parameters.R
+++ b/R/parameters.R
@@ -632,5 +632,6 @@ set_parameter_draw <- function(parameters, draw){
   for (name in names(parameter_draw)) {
     parameters[[name]] <- parameter_draw[[name]]
   }
+  parameters$parameter_draw <- draw
   return(parameters)
 }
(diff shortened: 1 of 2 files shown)


tbreweric:
> I know this is just updating the parameter set, but do you think it could be worth also passing this through to the simulation outputs to be rendered in the same way as repetition number, for example? Or maybe adding a switch that allows that to be rendered...? Just a thought! Otherwise looks great!

I know that we've talked often about reducing rather than increasing the number of rendered columns but I could see this being quite useful as it's a post-processing step I basically always do when running simulations with parameter variation. Thoughts @giovannic / @pwinskill?

tbreweric:
> Great, please add this side effect to the documentation for `set_parameter_draw`. Thanks!

I have amended the documentation for `set_parameter_draw` and the vignette that describes it to capture the side effect. I have also added an entry to the `get_parameters` documentation to explain the new parameter that gets added.

Before merging the changes, I'll wait and see if there is any feedback on @RJSheppard's suggestion about rendering the parameter draw in the simulation output.

PR #367, opened 2025-06-23
Enable users to run_simulations_with_reptitions() with correlated interventions

tbreweric:
The existing version of `run_simulations_with_reptitions()` makes it difficult/impossible to run simulations with repetitions. This would be a useful feature, so I have:
 
- Replaced the `overrides()`  argument with `parameters` & `correlations`
- Set default value of `correlations` to `NULL`
- Replaced the `run_simulations()` argument assignment in the function to accept these two new arguments
- Written two unit tests that confirm that the function runs successfully with and without correlations input

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -353,13 +353,15 @@ run_metapop_simulation <- function(
 #'
 #' @param timesteps the number of timesteps to run the simulation for
 #' @param repetitions n times to run the simulation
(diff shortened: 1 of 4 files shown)


tbreweric:
Seems we're failing checks - will fix and push changes.

giovannic (approved):
LGTM thanks!

RJSheppard (approved):

PR #377, opened 2025-08-06
Consistent EIR definition in the Pf/Pv models

RJSheppard:
Previously, vivax and falciparum calculations to generate expected bites from EIR were different, where the EIR of falciparum was at the level of the adult, whereas vivax was at the level of the whole population (taking into account fewer bites on children). For consistency, we now remove this difference, opting for the falciparum version. This will not make a difference to outputs, but changes the scaling for EIR to prevalence in the vivax model (EIR is abstract in the vivax model). A similar change is made in the vivax equilibrium package.


RJSheppard:
I was trying to adjust MW's model to match the Pf model, while still allowing for use of MW's original EIR definition for comparison if needed. We could decide it isn't important to preserve both methods, though.

Completely agree about input/output definitions - that tends to trip up new users!

I'm not sure what you mean by the changes in compatibility.R. For the EIR to have the same definition, it just needs to be treated the same way in both models...?

giovannic (approved):
~Love this, it's very clear. Thank you!~

giovannic (changes requested):
Will call for more clarification

